PR #2, opened 2018-07-16
[MDEP-568] respect excludeGroupIds in go-offline

cosmin:
instead of relying on "requiresDependencyResolution" to download artifacts, download the artifacts in the plugin code after filtering them with the provided parameters.

### DIFF
--- a/src/it/projects/mdep-568-go-offline-exclude/invoker.properties
+++ b/src/it/projects/mdep-568-go-offline-exclude/invoker.properties
@@ -1,0 +1,18 @@
+# Licensed to the Apache Software Foundation (ASF) under one
+# or more contributor license agreements.  See the NOTICE file
+# distributed with this work for additional information
+# regarding copyright ownership.  The ASF licenses this file
+# to you under the Apache License, Version 2.0 (the
+# "License"); you may not use this file except in compliance
+# with the License.  You may obtain a copy of the License at
+# 
+#   http://www.apache.org/licenses/LICENSE-2.0
+# 
+# Unless required by applicable law or agreed to in writing,
+# software distributed under the License is distributed on an
+# "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+# KIND, either express or implied.  See the License for the
+# specific language governing permissions and limitations
+# under the License.
+
+invoker.goals = ${project.groupId}:${project.artifactId}:${project.version}:go-offline
--- a/src/it/projects/mdep-568-go-offline-exclude/pom.xml
+++ b/src/it/projects/mdep-568-go-offline-exclude/pom.xml
@@ -1,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one
+  ~ or more contributor license agreements.  See the NOTICE file
+  ~ distributed with this work for additional information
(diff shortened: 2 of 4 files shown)


famod:
@hboutemy / @rfscholte / @khmarbaise is there anything blocking this from being merged?
LGTM, passes locally on Win 10 with Maven 3.6.0 and "Java version: 1.8.0_192, vendor: Oracle Corporation" as well as "Java version: 11.0.1, vendor: AdoptOpenJDK".

hboutemy:
sorry, I was busy on many other topics: this PR looks interesting

if I read the IT, there is no check: that's surprising (and limits the risk of failure: I didn't even try to run the IT without applying the patch to the mojo)

I have a few additional question about the feature:
- did it ever work?
- does the patch fix both dependencies and plugins/reports?
- for my understanding, what's the typical use case for excluding something, please?

cosmin:
It worked for me and I ended up having to publish a patched a version under `con.offbytwo` so I can use it in my builds. For me the use case is to exclude inter-module dependencies in a multi-module project (since the module artifacts can’t be downloaded from anywhere else but rather need to be built). My specific reason for wanting to do this is to cache all dependies needed to build a project (keyed by pom hashes) and use that cache to speed up future builds. This is the recommended workflow in CircleCI for example, but any other container based builds would want a similar workflow. 

I believe there was some discussion about this on the Jira ticket as well.

Theoderich:
Original Author of the patch here.

As @cosmin has said, this feature is needed to build a clean maven repository with all transitive dependencies of a project. In container based build engines like GitlabCI and CircleCi you need to copy around the repository between build containers. You want the repository to be as complete as possible so build steps don't need to download dependencies again and again, but also as small as possible so you don't need to move around to much data.

dependency:go-offline seemed to be the right tool. But it always tried to download the inter-project dependencies of my multi-module project that
a) Are not published anywhere, so the plugin fails
b) I wouldn't want to have in my clean repository anyway

With the changes of this patch, I was able to exclude the groupId if my project and get the required results.

So to answer your questions:
- did it ever work?
Not sure if it worked in any previous versions of the dependency plugin. The documentation at least suggests that it should work.

- does the patch fix both dependencies and plugins/reports?
Dependencies already work in the master version since they are not collected automatically via the requiresDependencyCollection annotation. Not sure about reports though.

- for my understanding, what's the typical use case for excluding something, please?
As described: Container based build platforms

In the meanttime, I have also published a plugin especially designed for downloading dependencies for build containers: https://github.com/qaware/go-offline-maven-plugin . It covers more edge cases than the patch proposed here, since it is a complete re-write.

ehrmann:
Any updates on this change?

andreysaksonov:
is something blocking it from being merged?

aiannucci:
Are you considering merging this change for an upcoming release? Thanks!

MarsianMan:
> * for my understanding, what's the typical use case for excluding something, please?

I have a multi-module project where Module B depends on Module A. `mvn dependency:go-offline` does not understand the intra-module dependency. Currently neither excluding group nor artifact ID work.

kgalieva:
Any updates on this change?

bmarwell:
Hi, the code base changed considerably. I created https://github.com/apache/maven-dependency-plugin/pull/135. @slachiewicz please review there.

jira-importer:
Resolve #1004

PR #13, opened 2019-04-27
[MDEP-651] - allow unpack dependency to configure entries overwrite

viqueen:
https://issues.apache.org/jira/browse/MDEP-651

Extending unpack-dependencies with an option to not overwrite archive entries with duplicate names. The latter can be problematic on case insensitive file systems.

From what I can tell, we always overwrite when unarchiving artifacts , therefore maintaining that to be the default behaviour.

This change will also require the following updates to plexus-archiver https://github.com/codehaus-plexus/plexus-archiver/pull/112

       pulling in other changes.
       commit message.

### DIFF
--- a/src/main/java/org/apache/maven/plugins/dependency/fromDependencies/UnpackDependenciesMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/fromDependencies/UnpackDependenciesMojo.java
@@ -83,12 +83,18 @@ public class UnpackDependenciesMojo
     private FileMapper[] fileMappers;
 
     /**
+     * @since 3.1.2
+     */
+    @Parameter( property = "mdep.unpack.overwrite", defaultValue = "true" )
+    private boolean overwrite;
+
+    /**
      * Main entry into mojo. This method gets the dependencies and iterates through each one passing it to
      * DependencyUtil.unpackFile().
      *
      * @throws MojoExecutionException with a message if an error occurs.
      * @see #getDependencySets(boolean)
-     * @see #unpack(Artifact, File, String)
+     * @see #unpack(Artifact, String, File, String, String, String, FileMapper[], boolean)
      */
     @Override
     protected void doExecute()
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/maven/plugins/dependency/fromDependencies/UnpackDependenciesMojo.java
```diff
@@ -86,7 +86,7 @@
      * @since 3.1.2
      */
     @Parameter( property = "mdep.unpack.overwrite", defaultValue = "true" )
-    private boolean overwrite;
+    private boolean overwrite = true;
```
khmarbaise: This is already don via the `defaultValue` from the annotation above.

khmarbaise:
It would be great having a unit/integration tests to check the behaviour...

hazendaz:
need this one badly.  Recent plexus archivers fail hard so we have to use 4.2.2 so that it does not.  even without tests its a problem currently if overwrite as documented all over doesn't actually work.  what can we do to get this going?

hazendaz:
see https://maven.apache.org/plugins/maven-dependency-plugin/usage.html, the documentation as is states this works, this shows that it does not, proven does not.

plexus archiver 4.2.2 simply states [WARNING] Unable to expand to file .... when it has issue.  All newer including 4.7.1 of plexus archiver blow up as file is open (ie windows).

hazendaz:
ok guess this may be different issue.  and much worse.  'unpack' isn't respecting this.  'unpack-dependencies' didn't have it.  willing to fix here but I doubt this even works.  Its simply to test.  I unzip chrome to .m2 and then open a copy, crashes everytime as indicated with newer plexus archiver but was a warning previously but really don't want it to re-extract anything.

hazendaz:
Adding a comment here to old comments I made.  The issue turned out to be the marker file location used by the dependency plugin.  Its default sits in target so 'clean' wipes that every time.  Moving that to m2 location instead resolved the issue for us where we share a live cache for running jenkins jobs on same boxes.  So we no longer get the behaviour we were seeing at that time and nothing needed to change other than moving where the marker directory was located.

See https://maven.apache.org/plugins/maven-dependency-plugin/unpack-mojo.html#markersDirectory

jira-importer:
Resolve #1082

PR #20, opened 2019-09-02
[MDEP-204] Rewrite go-offline goal w/ integration test

mthmulders:
Following this checklist to help us incorporate your contribution quickly and easily:

 - [X] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MDEP) filed for the change (usually before you start working on it).  Trivial changes like typos do not require a JIRA issue.  Your pull request should address just this issue, without pulling in other changes.
 - [X] Format the pull request title like `[MDEP-XXX] - Fixes bug in ApproximateQuantiles`, where you replace `MDEP-XXX` with the appropriate JIRA issue. Best practice is to use the JIRA issue title in the pull request title and in the first line of the commit message.
 - [X] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will be performed on your pull request automatically.

If your pull request is about ~20 lines of code you don't need to sign an [Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) if you are unsure please ask on the developers list.

To make clear that you license your contribution under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0) you have to acknowledge this by using the following check-box.

 - [X] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

This pull request fixes [MDEP-204](https://issues.apache.org/jira/browse/MDEP-204), where the `go-offline` goal did not work for multi-module projects with modules depending on other modules in the reactor. This was caused by the fact that it used Mavens internal dependency resolution mechanism without filtering out the artifacts that are already in the reactor. I've added a custom filter for that, and the plugin now only resolves those artifacts that the filter didn't remove.


src/test/java/org/apache/maven/plugins/dependency/CapturingLog.java
```diff
@@ -0,0 +1,179 @@
+package org.apache.maven.plugins.dependency;
+
+import org.apache.maven.plugin.logging.Log;
```
rfscholte: I would expect the rat-plugin to complain about this. It should be

1. package
1. license
1. imports

src/test/java/org/apache/maven/plugins/dependency/CapturingLog.java
```diff
@@ -0,0 +1,179 @@
...
+
+    /** {@inheritDoc} */
+    public void error( CharSequence content, Throwable error )
+    {
+        StringWriter sWriter = new StringWriter();
+        PrintWriter pWriter = new PrintWriter( sWriter );
+
+        error.printStackTrace( pWriter );
+
+        System.err.println( "[error] " + content.toString() + System.lineSeparator() + System.lineSeparator() + sWriter.toString() );
```
rfscholte: I know this code is has been moved, but System.err / System.out are evil.
Makes we wonder if we really need this class, it is not really capturing logs.
How about replacing its usage with a (Mockito) Mock? And simply delete this class?

src/main/java/org/apache/maven/plugins/dependency/resolvers/ExcludeReactorProjectsArtifactFilter.java
```diff
@@ -0,0 +1,95 @@
...
+                }
+            }
+        }
+
+        return results;
+    }
+
+    private boolean isArtifactInReactor( final Artifact artifact )
+    {
+        for ( final Artifact reactorArtifact : reactorArtifacts )
```
rfscholte: If reactorArtifacts are only used here, I'd suggest to change the constructor and field a bit, so we only have to transform once:

    private final Set<String> reactorArtifactsKeys;

Let constructor fill this field, so the code of this method can simply do a contains.

src/test/java/org/apache/maven/plugins/dependency/resolvers/ExcludeReactorProjectsArtifactFilterTest.java
```diff
@@ -0,0 +1,88 @@
+package org.apache.maven.plugins.dependency.resolvers;
+
+import org.apache.maven.artifact.Artifact;
```
rfscholte: Move license here

src/main/java/org/apache/maven/plugins/dependency/resolvers/ExcludeReactorProjectsArtifactFilter.java
mthmulders: Good catch!

src/test/java/org/apache/maven/plugins/dependency/CapturingLog.java
mthmulders: Yes, they're evil. I suspect that the original idea was that if there's an error, you want to see it, even in a test scenario. The plugin currently doesn't have a dependency on Mockito or the like, should we add it just for this case?

rfscholte: Yes please, it might very well be that the code was written before mocking frameworks existed.

rfscholte:
Merged, thanks for the patch!

jira-importer:
Resolve #1186

PR #27, opened 2019-11-24
Use Java 7 features

turbanoff:
collapse identical catch clauses
use generics where possible
fix typos
replace deprecated method usages
use diamond operator


rfscholte:
thanks!

PR #36, opened 2020-03-03
Replace more deprecated and unused code

elharo:
@eolivelli

### DIFF
--- a/src/test/java/org/apache/maven/plugins/dependency/resolvers/ExcludeReactorProjectsDependencyFilterTest.java
+++ b/src/test/java/org/apache/maven/plugins/dependency/resolvers/ExcludeReactorProjectsDependencyFilterTest.java
@@ -34,9 +34,6 @@ import java.util.HashSet;
 import java.util.Set;
 
 import static java.util.Collections.singletonList;
-import static org.hamcrest.CoreMatchers.is;
-import static org.hamcrest.CoreMatchers.containsString;
-import static org.junit.Assert.assertThat;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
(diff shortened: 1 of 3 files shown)


src/test/java/org/apache/maven/plugins/dependency/resolvers/ExcludeReactorProjectsDependencyFilterTest.java
```diff
@@ -124,7 +119,7 @@ public Dependency getDependency() {
 
         ArgumentCaptor<String> captor = ArgumentCaptor.forClass( String.class );
         verify( log ).debug( captor.capture() );
-        assertThat( captor.getValue(), containsString( "Skipped dependency" ) );
+        assertTrue( captor.getValue().contains( "Skipped dependency" ) );
```
Tibor17: revert This.
`assertTrue` and the old JUnit assertios are the bad practice.

src/test/java/org/apache/maven/plugins/dependency/resolvers/ExcludeReactorProjectsDependencyFilterTest.java
```diff
@@ -163,8 +158,6 @@ public Dependency getDependency() {
             }
         };
 
-        final boolean result = filter.accept( node , Collections.<Node>emptyList() );
-
-        assertThat( result, is( true ));
+        assertTrue( filter.accept( node , Collections.<Node>emptyList() ) );
```
Tibor17: Here as well and use AssertJ instead of Java Hamcrest.

elharo: This is standard JUnit, not Hamcrest. AssertJ would add an extra dependency we don't currently use or need.

src/test/java/org/apache/maven/plugins/dependency/resolvers/ExcludeReactorProjectsDependencyFilterTest.java
elharo: Reference?

PR #40, opened 2020-03-28
update commons-lang

elharo:
@michael-o


elharo:
Hmm, looks like they did change the package? So how are the tests passing?

https://builds.apache.org/job/maven-box/job/maven-dependency-plugin/job/lang/

Must be a transitive dependency somewhere. Fixed.

PR #44, opened 2020-04-02
[MDEP-676] fix test

elharo:
@hboutemy the upgrade broke one of the integration tests. Fixing this is going to require moving to Maven 3.1 as a minimum which requires some other changes.


jira-importer:
Resolve #1185

PR #50, opened 2020-05-06
MDEP-648: Add location of listed repository.

pmoerenhout:
After maven-artifact-transfer 0.13.0 has been released, the PR will use the new functions from the maven-artifact-transfer to resolve the used repositories, and in verbose mode, also show the (local) location where this repository is defined.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -86,6 +86,9 @@ under the License.
     <contributor>
       <name>Maarten Mulders</name>
     </contributor>
+    <contributor>
+      <name>Pim Moerenhout</name>
+    </contributor>
   </contributors>
 
   <properties>
@@ -243,7 +246,7 @@ under the License.
     <dependency>
       <groupId>org.apache.maven.shared</groupId>
       <artifactId>maven-artifact-transfer</artifactId>
-      <version>0.11.0</version>
+      <version>0.13.0</version>
     </dependency>
     <dependency>
       <groupId>org.apache.maven.shared</groupId>
@@ -254,7 +257,7 @@ under the License.
     <dependency>
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
-      <version>3.6</version>
+      <version>3.8.1</version>
     </dependency>
 
     <dependency>
@@ -317,7 +320,7 @@ under the License.
     <dependency>
       <groupId>org.apache.maven.wagon</groupId>
       <artifactId>wagon-http-lightweight</artifactId>
-      <version>3.3.4</version>
+      <version>3.4.0</version>
       <scope>provided</scope>
     </dependency>
 
--- a/src/main/java/org/apache/maven/plugins/dependency/resolvers/ListRepositoriesMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/resolvers/ListRepositoriesMojo.java
@@ -19,15 +19,36 @@ package org.apache.maven.plugins.dependency.resolvers;
  * under the License.
  */
 
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.apache.maven.artifact.Artifact;
+import org.apache.maven.artifact.ArtifactUtils;
 import org.apache.maven.artifact.repository.ArtifactRepository;
+import org.apache.maven.model.Model;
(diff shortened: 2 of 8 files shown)


pom.xml
```diff
@@ -254,7 +257,7 @@ under the License.
     <dependency>
       <groupId>org.apache.commons</groupId>
       <artifactId>commons-lang3</artifactId>
-      <version>3.6</version>
```
mthmulders, line 257: Is this related to the functionality that you've implemented?

pom.xml
```diff
@@ -317,7 +320,7 @@ under the License.
     <dependency>
       <groupId>org.apache.maven.wagon</groupId>
       <artifactId>wagon-http-lightweight</artifactId>
-      <version>3.3.4</version>
```
mthmulders, line 320: Is this related to the functionality that you've implemented?

src/main/java/org/apache/maven/plugins/dependency/resolvers/ListRepositoriesMojo.java
```diff
@@ -40,12 +61,46 @@
...
+     * Remote repositories used for the project.
+     */
+    @Parameter( defaultValue = "${project.remoteArtifactRepositories}", required = true, readonly = true )
+    private List<ArtifactRepository> remoteRepositories;
+
+    /**
+     * Sets whether the plugin runs in verbose mode. As of plugin version 2.3, the default value is derived from Maven's
+     * global debug flag (compare command line switch <code>-X</code>). <br/>
+     *
+     * @since 3.1.2
```
mthmulders, line 99: The `@since` contradicts the first line of the Javadoc.

src/site/apt/usage.apt.vm
```diff
@@ -687,6 +687,17 @@ mvn dependency:build-classpath -Dmdep.outputFile=cp.txt
   This goal is used to list all the repositories that this build depends upon. It will show repositories defined in your settings, 
   poms and declared in transitive dependency poms.
 
+  This goal can be executed from the command line:
```
mthmulders, line 690: Good to see you updated the documentation as well!

src/main/java/org/apache/maven/plugins/dependency/resolvers/ListRepositoriesMojo.java
```diff
@@ -55,22 +110,276 @@
...
+        sb.append( repository.getId() );
+        sb.append( " (" );
+        sb.append( repository.getUrl() );
+        sb.append( ")" );
+        return sb.toString();
+    }
+
+    private boolean isVerbose()
+    {
+        return ( verbose || getLog().isDebugEnabled() );
```
mthmulders, line 367: It's a bit strange that `isVerbose` checks for `log.isDebugEnabled()`, while there is a `verbose` flag already which is triggered (according to its Javadoc) by the `-X` flag in the CLI.

src/main/java/org/apache/maven/plugins/dependency/resolvers/ListRepositoriesMojo.java
```diff
@@ -55,22 +110,276 @@
...
+        if ( parent != null )
+        {
+            Model originalModel = parent.getOriginalModel();
+            if ( originalModel.getRepositories().size() != 0
+                || originalModel.getPluginRepositories().size() != 0 )
             {
-                this.getLog().info( repo.toString() );
+                String artifactKey =
+                    ArtifactUtils.key( parent.getGroupId(), parent.getArtifactId(), parent.getVersion() );
+                MavenProject parentPom = getMavenProject( artifactKey );
```
mthmulders, line 328: It seems to me as if you first concatenate _groupId_, _artifactId_ and _version_ into an _artifactKey_ (line 326), only to pass it to `getMavenProject(String)` which then uses `getArtifactCoordinate` to disassemble then into their original values. Why is that??

slawekjaranowski:
@pmoerenhout thanks also for efforts on it ... we should not use maven-artifact-transfer at all
It should be reimplemented with using direct resolver API like I did in https://github.com/apache/maven-dependency-plugin/pull/301

So I close it now.

jira-importer:
Resolve #1081

PR #53, opened 2020-05-18
Moved GetMojo logic to an abstract class. This allows new mojos to im…

jonvolfson:
…plement functionality on top of checking if an artifact can be retrieved remotely.

This will allow the logic in the old GetMojo to be used for [MDEP-645].

### DIFF
--- a/src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
@@ -1,0 +1,396 @@
+package org.apache.maven.plugins.dependency;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Enumeration;
+import java.util.List;
+import java.util.Map;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.apache.maven.artifact.handler.ArtifactHandler;
+import org.apache.maven.artifact.handler.manager.ArtifactHandlerManager;
+import org.apache.maven.artifact.repository.ArtifactRepository;
+import org.apache.maven.artifact.repository.ArtifactRepositoryPolicy;
+import org.apache.maven.artifact.repository.MavenArtifactRepository;
+import org.apache.maven.artifact.repository.layout.ArtifactRepositoryLayout;
+import org.apache.maven.execution.MavenSession;
+import org.apache.maven.plugin.AbstractMojo;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.MojoFailureException;
+import org.apache.maven.plugins.annotations.Component;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.project.DefaultProjectBuildingRequest;
+import org.apache.maven.project.ProjectBuildingRequest;
+import org.apache.maven.repository.RepositorySystem;
+import org.apache.maven.settings.Settings;
+import org.apache.maven.shared.transfer.artifact.ArtifactCoordinate;
+import org.apache.maven.shared.transfer.artifact.DefaultArtifactCoordinate;
+import org.apache.maven.shared.transfer.artifact.resolve.ArtifactResolver;
+import org.apache.maven.shared.transfer.artifact.resolve.ArtifactResolverException;
+import org.apache.maven.shared.transfer.artifact.resolve.ArtifactResult;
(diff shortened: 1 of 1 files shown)


jonvolfson:
@bimargulies @elharo

src/main/java/org/apache/maven/plugins/dependency/AbstractGetMojo.java
```diff
@@ -0,0 +1,355 @@
...
+import org.apache.maven.shared.transfer.dependencies.resolve.DependencyResolver;
+import org.apache.maven.shared.transfer.dependencies.resolve.DependencyResolverException;
+import org.codehaus.plexus.util.StringUtils;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
```
bimargulies-google, line 53: Add javadoc comment here that explains what this class is good for.

src/main/java/org/apache/maven/plugins/dependency/AbstractGetMojo.java
```diff
@@ -0,0 +1,355 @@
...
+            throws MojoExecutionException, MojoFailureException;
+
+    @Override
+    public void execute()
+            throws MojoExecutionException, MojoFailureException
+    {
+        doExecute();
+    }
+
+    public ProjectBuildingRequest buildBuildingRequest()
```
bimargulies-google, line 170: Need javadoc comment on here, should explain that the expectation is that derived classes will call this.

src/main/java/org/apache/maven/plugins/dependency/AbstractGetMojo.java
```diff
@@ -0,0 +1,355 @@
...
+    /**
+     * Skip plugin execution completely.
+     *
+     * @since 2.7
+     */
+    @Parameter( property = "mdep.skip", defaultValue = "false" )
+    private boolean skip;
+
+    /**
+     * @throws MojoExecutionException {@link MojoExecutionException}
```
bimargulies-google, line 157: Needs text stating that each derived class will provide this.

src/main/java/org/apache/maven/plugins/dependency/GetMojo.java
```diff
@@ -58,292 +58,38 @@
...
-    private boolean skip;
 
     @Override
-    public void execute()
-        throws MojoExecutionException, MojoFailureException
+    protected void doExecute()
+            throws MojoExecutionException, MojoFailureException
     {
+
         if ( isSkip() )
```
bimargulies-google, line 70: Skip logic can be in base class.

src/main/java/org/apache/maven/plugins/dependency/AbstractGetMojo.java
jonvolfson: Resolved

src/main/java/org/apache/maven/plugins/dependency/AbstractGetMojo.java
jonvolfson: Resolved

src/main/java/org/apache/maven/plugins/dependency/AbstractGetMojo.java
jonvolfson: Resolved

src/main/java/org/apache/maven/plugins/dependency/GetMojo.java
jonvolfson: Resolved

src/main/java/org/apache/maven/plugins/dependency/AbstractGetMojo.java
```diff
@@ -0,0 +1,367 @@
...
+ */
+public abstract class AbstractGetMojo
+        extends AbstractMojo
+{
+    private static final Pattern ALT_REPO_SYNTAX_PATTERN = Pattern.compile( "(.+)::(.*)::(.+)" );
+
+    @Parameter( defaultValue = "${session}", required = true, readonly = true )
+    private MavenSession session;
+
+    /**
```
elharo, line 65: no empty doc comments please

src/main/java/org/apache/maven/plugins/dependency/AbstractGetMojo.java
```diff
@@ -0,0 +1,367 @@
...
+    private List<ArtifactRepository> pomRemoteRepositories;
+
+    /**
+     * Download transitively, retrieving the specified artifact and all of its dependencies.
+     */
+    @Parameter( property = "transitive", defaultValue = "true" )
+    private boolean transitive = true;
+
+    /**
+     * Skip plugin execution completely. Derived classes will provide this.
```
elharo, line 152: skip seems to be done in this class

src/main/java/org/apache/maven/plugins/dependency/AbstractGetMojo.java
jonvolfson: resolved

src/main/java/org/apache/maven/plugins/dependency/AbstractGetMojo.java
jonvolfson: resolved

src/main/java/org/apache/maven/plugins/dependency/AbstractGetMojo.java
```diff
@@ -0,0 +1,358 @@
...
+    private String classifier;
+
+    /**
+     * The packaging of the artifact to download. Ignored if {@link #artifact} is used.
+     */
+    @Parameter( property = "packaging", defaultValue = "jar" )
+    private String packaging = "jar";
+
+    /**
+     * Repositories in the format id::[layout]::url or just url, separated by comma. ie.
```
elharo, line 121: ie. --> For example,

src/main/java/org/apache/maven/plugins/dependency/AbstractGetMojo.java
```diff
@@ -0,0 +1,358 @@
...
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+/**
+ * Abstract parent class used by mojos that retrieve artifacts from the remote repository.
+ */
+public abstract class AbstractGetMojo
```
elharo, line 57: Is there maybe a better name for this class? GetMojo implies this is for the dependency:get goal.

PR #57, opened 2020-05-26
[MDEP-645] Created new mojo GetClassesMojo which lists all class dependencies for a specified artifact.

jonvolfson:
@elharo @bimargulies

### DIFF
--- a/src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
@@ -57,6 +57,10 @@ import org.apache.maven.shared.transfer.dependencies.resolve.DependencyResolver;
 import org.apache.maven.shared.transfer.dependencies.resolve.DependencyResolverException;
 import org.codehaus.plexus.util.StringUtils;
 
+
+/**
+ * Retrieves and lists all class dependencies for the specified artifact from the specified remote repositories.
+ */
 @Mojo( name = "get-classes", requiresProject = false, threadSafe = true )
 public class GetClassesMojo
     extends AbstractMojo
@@ -162,7 +166,8 @@ public class GetClassesMojo
             getLog().info( "Retrieving classes of dependencies for " + artifact );
             if ( isTransitive() )
             {
-                Iterable<ArtifactResult> artifacts = getDependencyResolver().resolveDependencies( buildingRequest, coordinate, null );
+                Iterable<ArtifactResult> artifacts = getDependencyResolver()
+                        .resolveDependencies( buildingRequest, coordinate, null );
 
                 for ( ArtifactResult result : artifacts )
                 {
@@ -171,7 +176,8 @@ public class GetClassesMojo
             }
             else
             {
-                ArtifactResult result = getArtifactResolver().resolveArtifact( buildingRequest, toArtifactCoordinate( coordinate ) );
+                ArtifactResult result = getArtifactResolver()
+                        .resolveArtifact( buildingRequest, toArtifactCoordinate( coordinate ) );
 
                 printClassesFromArtifactResult( result );
             }
@@ -190,11 +196,12 @@ public class GetClassesMojo
 
         while ( e.hasMoreElements() )
         {
-            JarEntry entry = ( JarEntry )e.nextElement();
+            JarEntry entry = (JarEntry) e.nextElement();
             String name = entry.getName();
 
             // filter out files that do not end in .class
-            if ( name.length() <= 6 || !name.substring( name.length() - 6 ).equals( ".class" )) {
+            if ( name.length() <= 6 || !name.substring( name.length() - 6 ).equals( ".class" ) )
+            {
                 continue;
             }
 
@@ -268,7 +275,7 @@ public class GetClassesMojo
         return buildingRequest;
     }
 
-    protected ArtifactCoordinate toArtifactCoordinate(DependableCoordinate dependableCoordinate )
+    protected ArtifactCoordinate toArtifactCoordinate( DependableCoordinate dependableCoordinate )
     {
         ArtifactHandler artifactHandler = artifactHandlerManager.getArtifactHandler( dependableCoordinate.getType() );
         DefaultArtifactCoordinate artifactCoordinate = new DefaultArtifactCoordinate();
@@ -352,7 +359,10 @@ public class GetClassesMojo
     /**
      * @param artifact The artifact
      */
-    public void setArtifact( String artifact ) { this.artifact = artifact; }
+    public void setArtifact( String artifact )
+    {
+        this.artifact = artifact;
+    }
 
     /**
      * @param groupId The groupId.
--- a/src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
@@ -166,14 +166,14 @@ public class GetClassesMojo
 
                 for ( ArtifactResult result : artifacts )
                 {
-                    PrintClassesFromArtifactResult( result );
+                    printClassesFromArtifactResult( result );
                 }
             }
             else
             {
                 ArtifactResult result = getArtifactResolver().resolveArtifact( buildingRequest, toArtifactCoordinate( coordinate ) );
 
-                PrintClassesFromArtifactResult( result );
+                printClassesFromArtifactResult( result );
             }
         }
         catch ( ArtifactResolverException | DependencyResolverException | IOException e )
@@ -182,7 +182,7 @@ public class GetClassesMojo
         }
     }
 
-    public void PrintClassesFromArtifactResult( ArtifactResult result )
+    public void printClassesFromArtifactResult( ArtifactResult result )
             throws IOException
     {
         JarFile jarFile = new JarFile( result.getArtifact().getFile() );
--- a/src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
@@ -1,0 +1,396 @@
+package org.apache.maven.plugins.dependency;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.Enumeration;
+import java.util.List;
+import java.util.Map;
+import java.util.jar.JarEntry;
+import java.util.jar.JarFile;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+
+import org.apache.maven.artifact.handler.ArtifactHandler;
+import org.apache.maven.artifact.handler.manager.ArtifactHandlerManager;
+import org.apache.maven.artifact.repository.ArtifactRepository;
+import org.apache.maven.artifact.repository.ArtifactRepositoryPolicy;
+import org.apache.maven.artifact.repository.MavenArtifactRepository;
+import org.apache.maven.artifact.repository.layout.ArtifactRepositoryLayout;
+import org.apache.maven.execution.MavenSession;
+import org.apache.maven.plugin.AbstractMojo;
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.MojoFailureException;
+import org.apache.maven.plugins.annotations.Component;
+import org.apache.maven.plugins.annotations.Mojo;
+import org.apache.maven.plugins.annotations.Parameter;
+import org.apache.maven.project.DefaultProjectBuildingRequest;
+import org.apache.maven.project.ProjectBuildingRequest;
+import org.apache.maven.repository.RepositorySystem;
+import org.apache.maven.settings.Settings;
+import org.apache.maven.shared.transfer.artifact.ArtifactCoordinate;
+import org.apache.maven.shared.transfer.artifact.DefaultArtifactCoordinate;
+import org.apache.maven.shared.transfer.artifact.resolve.ArtifactResolver;
+import org.apache.maven.shared.transfer.artifact.resolve.ArtifactResolverException;
+import org.apache.maven.shared.transfer.artifact.resolve.ArtifactResult;
+import org.apache.maven.shared.transfer.dependencies.DefaultDependableCoordinate;
+import org.apache.maven.shared.transfer.dependencies.DependableCoordinate;
+import org.apache.maven.shared.transfer.dependencies.resolve.DependencyResolver;
+import org.apache.maven.shared.transfer.dependencies.resolve.DependencyResolverException;
+import org.codehaus.plexus.util.StringUtils;
+
+@Mojo( name = "get-classes", requiresProject = false, threadSafe = true )
+public class GetClassesMojo
+    extends AbstractMojo
+{
+    private static final Pattern ALT_REPO_SYNTAX_PATTERN = Pattern.compile( "(.+)::(.*)::(.+)" );
+
+    @Parameter( defaultValue = "${session}", required = true, readonly = true )
+    private MavenSession session;
+
+    @Component
+    private ArtifactResolver artifactResolver;
+
+    @Component
+    private DependencyResolver dependencyResolver;
+
+    @Component
+    private ArtifactHandlerManager artifactHandlerManager;
+
+    /**
+     * Map that contains the layouts.
+     */
+    @Component( role = ArtifactRepositoryLayout.class )
+    private Map<String, ArtifactRepositoryLayout> repositoryLayouts;
+
+    /**
+     * The repository system.
+     */
+    @Component
+    private RepositorySystem repositorySystem;
+
+    private DefaultDependableCoordinate coordinate = new DefaultDependableCoordinate();
+
+    /**
+     * The groupId of the artifact to download. Ignored if {@link #artifact} is used.
+     */
+    @Parameter( property = "groupId" )
+    private String groupId;
+
+    /**
+     * The artifactId of the artifact to download. Ignored if {@link #artifact} is used.
+     */
+    @Parameter( property = "artifactId" )
+    private String artifactId;
+
+    /**
+     * The version of the artifact to download. Ignored if {@link #artifact} is used.
+     */
+    @Parameter( property = "version" )
+    private String version;
+
+    /**
+     * The classifier of the artifact to download. Ignored if {@link #artifact} is used.
+     *
+     * @since 2.3
+     */
+    @Parameter( property = "classifier" )
+    private String classifier;
+
+    /**
+     * The packaging of the artifact to download. Ignored if {@link #artifact} is used.
+     */
+    @Parameter( property = "packaging", defaultValue = "jar" )
+    private String packaging = "jar";
+
+    /**
+     * Repositories in the format id::[layout]::url or just url, separated by comma. ie.
+     * central::default::https://repo.maven.apache.org/maven2,myrepo::::https://repo.acme.com,https://repo.acme2.com
+     */
+    @Parameter( property = "remoteRepositories" )
+    private String remoteRepositories;
+
+    /**
+     * A string of the form groupId:artifactId:version[:packaging[:classifier]].
+     */
+    @Parameter( property = "artifact" )
+    private String artifact;
+
+    @Parameter( defaultValue = "${project.remoteArtifactRepositories}", readonly = true, required = true )
+    private List<ArtifactRepository> pomRemoteRepositories;
+
+    /**
+     * Download transitively, retrieving the specified artifact and all of its dependencies.
+     */
+    @Parameter( property = "transitive", defaultValue = "true" )
+    private boolean transitive = true;
+
+    /**
+     * Skip plugin execution completely.
+     *
+     * @since 2.7
+     */
+    @Parameter( property = "mdep.skip", defaultValue = "false" )
+    private boolean skip;
+
+    @Override
+    public void execute() throws MojoExecutionException, MojoFailureException
+    {
+        ProjectBuildingRequest buildingRequest = buildBuildingRequest();
+        DefaultDependableCoordinate coordinate = getCoordinate();
+
+        try
+        {
+            getLog().info( "Retrieving classes of dependencies for " + artifact );
+            if ( isTransitive() )
+            {
+                Iterable<ArtifactResult> artifacts = getDependencyResolver().resolveDependencies( buildingRequest, coordinate, null );
+
+                for ( ArtifactResult result : artifacts )
+                {
+                    PrintClassesFromArtifactResult( result );
+                }
+            }
+            else
+            {
+                ArtifactResult result = getArtifactResolver().resolveArtifact( buildingRequest, toArtifactCoordinate( coordinate ) );
+
+                PrintClassesFromArtifactResult( result );
+            }
+        }
+        catch ( ArtifactResolverException | DependencyResolverException | IOException e )
+        {
+            throw new MojoExecutionException( "Couldn't download artifact: " + e.getMessage(), e );
+        }
+    }
+
+    public void PrintClassesFromArtifactResult( ArtifactResult result )
+            throws IOException
+    {
+        JarFile jarFile = new JarFile( result.getArtifact().getFile() );
+        Enumeration e = jarFile.entries();
+
+        while ( e.hasMoreElements() )
+        {
+            JarEntry entry = ( JarEntry )e.nextElement();
+            String name = entry.getName();
+
+            // filter out files that do not end in .class
+            if ( name.length() <= 6 || !name.substring( name.length() - 6 ).equals( ".class" )) {
+                continue;
+            }
+
+            // remove .class from the end and change format to use periods instead of forward slashes
+            name = name.substring( 0, name.length() - 6 ).replace( '/', '.' );
+            getLog().info( name );
+        }
+        jarFile.close();
+    }
+
+    public ProjectBuildingRequest buildBuildingRequest()
+            throws MojoExecutionException, MojoFailureException
+    {
+        if ( coordinate.getArtifactId() == null && artifact == null )
+        {
+            throw new MojoFailureException( "You must specify an artifact, "
+                    + "e.g. -Dartifact=org.apache.maven.plugins:maven-downloader-plugin:1.0" );
+        }
+        if ( artifact != null )
+        {
+            String[] tokens = StringUtils.split( artifact, ":" );
+            if ( tokens.length < 3 || tokens.length > 5 )
+            {
+                throw new MojoFailureException( "Invalid artifact, you must specify "
+                        + "groupId:artifactId:version[:packaging[:classifier]] " + artifact );
+            }
+            coordinate.setGroupId( tokens[0] );
+            coordinate.setArtifactId( tokens[1] );
+            coordinate.setVersion( tokens[2] );
+            if ( tokens.length >= 4 )
+            {
+                coordinate.setType( tokens[3] );
+            }
+            if ( tokens.length == 5 )
+            {
+                coordinate.setClassifier( tokens[4] );
+            }
+        }
+
+        ArtifactRepositoryPolicy always =
+                new ArtifactRepositoryPolicy( true, ArtifactRepositoryPolicy.UPDATE_POLICY_ALWAYS,
+                        ArtifactRepositoryPolicy.CHECKSUM_POLICY_WARN );
+
+        List<ArtifactRepository> repoList = new ArrayList<>();
+
+        if ( pomRemoteRepositories != null )
+        {
+            repoList.addAll( pomRemoteRepositories );
+        }
+
+        if ( remoteRepositories != null )
+        {
+            // Use the same format as in the deploy plugin id::layout::url
+            String[] repos = StringUtils.split( remoteRepositories, "," );
+            for ( String repo : repos )
+            {
+                repoList.add( parseRepository( repo, always ) );
+            }
+        }
+
+        ProjectBuildingRequest buildingRequest =
+                new DefaultProjectBuildingRequest( session.getProjectBuildingRequest() );
+
+        Settings settings = session.getSettings();
+        repositorySystem.injectMirror( repoList, settings.getMirrors() );
+        repositorySystem.injectProxy( repoList, settings.getProxies() );
+        repositorySystem.injectAuthentication( repoList, settings.getServers() );
+
+        buildingRequest.setRemoteRepositories( repoList );
+
+        return buildingRequest;
+    }
+
+    protected ArtifactCoordinate toArtifactCoordinate(DependableCoordinate dependableCoordinate )
+    {
+        ArtifactHandler artifactHandler = artifactHandlerManager.getArtifactHandler( dependableCoordinate.getType() );
+        DefaultArtifactCoordinate artifactCoordinate = new DefaultArtifactCoordinate();
+        artifactCoordinate.setGroupId( dependableCoordinate.getGroupId() );
+        artifactCoordinate.setArtifactId( dependableCoordinate.getArtifactId() );
+        artifactCoordinate.setVersion( dependableCoordinate.getVersion() );
+        artifactCoordinate.setClassifier( dependableCoordinate.getClassifier() );
+        artifactCoordinate.setExtension( artifactHandler.getExtension() );
+        return artifactCoordinate;
+    }
+
+    ArtifactRepository parseRepository( String repo, ArtifactRepositoryPolicy policy )
+            throws MojoFailureException
+    {
+        // if it's a simple url
+        String id = "temp";
+        ArtifactRepositoryLayout layout = getLayout( "default" );
+        String url = repo;
+
+        // if it's an extended repo URL of the form id::layout::url
+        if ( repo.contains( "::" ) )
+        {
+            Matcher matcher = ALT_REPO_SYNTAX_PATTERN.matcher( repo );
+            if ( !matcher.matches() )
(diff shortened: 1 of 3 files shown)


khmarbaise:
First please create an appropriate JIRA issue for that and describe the use case for this.

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,396 @@
...
+     */
+    @Parameter( property = "transitive", defaultValue = "true" )
+    private boolean transitive = true;
+
+    /**
+     * Skip plugin execution completely.
+     *
+     * @since 2.7
+     */
+    @Parameter( property = "mdep.skip", defaultValue = "false" )
```
bimargulies-google, line 155: I think you can remove all of the 'skip' processing; it's not terribly useful for a mojo intended for CLI.

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,396 @@
...
+            name = name.substring( 0, name.length() - 6 ).replace( '/', '.' );
+            getLog().info( name );
+        }
+        jarFile.close();
+    }
+
+    public ProjectBuildingRequest buildBuildingRequest()
+            throws MojoExecutionException, MojoFailureException
+    {
+        if ( coordinate.getArtifactId() == null && artifact == null )
```
bimargulies-google, line 218: You don't use coordinate below. Are you missing an else clause? Is that the code that would run if someone used -DgroupId ... instead of -Dartifact?

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,396 @@
...
+        try
+        {
+            getLog().info( "Retrieving classes of dependencies for " + artifact );
+            if ( isTransitive() )
+            {
+                Iterable<ArtifactResult> artifacts = getDependencyResolver().resolveDependencies( buildingRequest, coordinate, null );
+
+                for ( ArtifactResult result : artifacts )
+                {
+                    printClassesFromArtifactResult( result );
```
bimargulies-google, line 174: Print the file name of the jar before printing its contents if you are printing more than one jar file's worth?

jonvolfson:
Updated PR title. This is a fix for MDEP-645

src/test/java/org/apache/maven/plugins/dependency/TestGetClassesMojo.java
```diff
@@ -0,0 +1,88 @@
...
+    public void testGetClassesNotTransitive()
+            throws Exception
+    {
+        setVariableValueToObject( mojo, "remoteRepositories", "central::default::https://repo.maven.apache.org/maven2,"
+                + "central::::https://repo.maven.apache.org/maven2," + "https://repo.maven.apache.org/maven2" );
+        mojo.setArtifact("org.apache.commons:commons-lang3:3.6");
+        setVariableValueToObject( mojo, "transitive", Boolean.FALSE );
+
+        mojo.execute();
+        return;
```
bimargulies-google, line 74: No need for return.

src/test/java/org/apache/maven/plugins/dependency/TestGetClassesMojo.java
```diff
@@ -0,0 +1,88 @@
...
+    public void testGetClassesTransitive()
+            throws Exception
+    {
+        setVariableValueToObject( mojo, "remoteRepositories", "central::default::https://repo.maven.apache.org/maven2,"
+                + "central::::https://repo.maven.apache.org/maven2," + "https://repo.maven.apache.org/maven2" );
+        mojo.setArtifact("org.apache.commons:commons-lang3:3.6");
+        setVariableValueToObject( mojo, "transitive", Boolean.TRUE );
+
+        mojo.execute();
+        return;
```
bimargulies-google, line 86: No need for return.

src/test/java/org/apache/maven/plugins/dependency/TestGetClassesMojo.java
```diff
@@ -0,0 +1,88 @@
...
+        setVariableValueToObject( mojo, "remoteRepositories", "central::default::https://repo.maven.apache.org/maven2,"
+                + "central::::https://repo.maven.apache.org/maven2," + "https://repo.maven.apache.org/maven2" );
+        mojo.setArtifact("org.apache.commons:commons-lang3:3.6");
+        setVariableValueToObject( mojo, "transitive", Boolean.FALSE );
+
+        mojo.execute();
+        return;
+    }
+
+    public void testGetClassesTransitive()
```
bimargulies-google, line 77: Um, this does not seem to test anything except that it does not explode.

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,396 @@
...
+            String name = entry.getName();
+
+            // filter out files that do not end in .class
+            if ( name.length() <= 6 || !name.substring( name.length() - 6 ).equals( ".class" )) {
+                continue;
+            }
+
+            // remove .class from the end and change format to use periods instead of forward slashes
+            name = name.substring( 0, name.length() - 6 ).replace( '/', '.' );
+            getLog().info( name );
```
bimargulies-google, line 210: I don't think that using the log is the best delivery mechanism, but I could be confused. @elharo what do you think? Also, I have no idea how to test for log output, you'd need to mock the log somehow.

src/test/java/org/apache/maven/plugins/dependency/TestGetClassesMojo.java
jonvolfson: When listing classes for something like org.apache.maven:maven-model:2.0.9, setting -Dtransitive=true lists more classes than the false alternative. It's possible that the artifact used in the test has little/no transitive dependencies which means the results would be similar/the same.

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
rmannibucau: +1, can be neat to be able to inject the list of classes in a maven project properties to reuse it downstream (exec-maven-plugin is my immediate thought but i'm sure there are more use cases). A dump in a file can be useful too.

side note: what about multi release jars (META-INF/versions)? should it be filtered?

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
rmannibucau: @bimargulies-google I'd keep it since you never know how it is used and maven does not control that

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
jonvolfson: Fixed. Removed the coordinate.getArtifactId() part. This was recycled code from the GetMojo with a similar use case.

src/test/java/org/apache/maven/plugins/dependency/TestGetClassesMojo.java
jonvolfson: Fixed.

src/test/java/org/apache/maven/plugins/dependency/TestGetClassesMojo.java
jonvolfson: Fixed.

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,406 @@
...
+import org.apache.maven.shared.transfer.artifact.resolve.ArtifactResult;
+import org.apache.maven.shared.transfer.dependencies.DefaultDependableCoordinate;
+import org.apache.maven.shared.transfer.dependencies.DependableCoordinate;
+import org.apache.maven.shared.transfer.dependencies.resolve.DependencyResolver;
+import org.apache.maven.shared.transfer.dependencies.resolve.DependencyResolverException;
+import org.codehaus.plexus.util.StringUtils;
+
+
+/**
+ * Retrieves and lists all class dependencies for the specified artifact from the specified remote repositories.
```
elharo, line 62: class dependencies for --> "classes contained in" as these are not dependencies

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,406 @@
...
+ */
+@Mojo( name = "get-classes", requiresProject = false, threadSafe = true )
+public class GetClassesMojo
```
elharo, line 65: Perhaps ListClassesMojo?

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,406 @@
...
+    /**
+     * The repository system.
+     */
+    @Component
+    private RepositorySystem repositorySystem;
+
+    private DefaultDependableCoordinate coordinate = new DefaultDependableCoordinate();
+
+    /**
+     * The groupId of the artifact to download. Ignored if {@link #artifact} is used.
```
elharo, line 97: group ID

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,406 @@
...
+     */
+    @Parameter( property = "groupId" )
+    private String groupId;
+
+    /**
+     * The artifactId of the artifact to download. Ignored if {@link #artifact} is used.
```
elharo, line 103: artifact ID

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,406 @@
...
+     * @since 2.3
+     */
+    @Parameter( property = "classifier" )
+    private String classifier;
+
+    /**
+     * The packaging of the artifact to download. Ignored if {@link #artifact} is used.
+     */
+    @Parameter( property = "packaging", defaultValue = "jar" )
+    private String packaging = "jar";
```
elharo, line 126: will this work for anything that's not a jar?

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,406 @@
...
+
+    /**
+     * Repositories in the format id::[layout]::url or just url, separated by comma. ie.
```
elharo, line 129: url, --> URLs,

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,406 @@
...
+     * A string of the form groupId:artifactId:version[:packaging[:classifier]].
+     */
+    @Parameter( property = "artifact" )
+    private String artifact;
+
+    @Parameter( defaultValue = "${project.remoteArtifactRepositories}", readonly = true, required = true )
+    private List<ArtifactRepository> pomRemoteRepositories;
+
+    /**
+     * Download transitively, retrieving the specified artifact and all of its dependencies.
```
elharo, line 145: I don't think this should follow transitive dependencies

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,406 @@
...
+     */
+    @Parameter( property = "transitive", defaultValue = "true" )
+    private boolean transitive = true;
+
+    /**
+     * Skip plugin execution completely.
+     *
+     * @since 2.7
```
elharo, line 153: `@since` doesn't make sense here

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,406 @@
...
+
+    @Override
+    public void execute() throws MojoExecutionException, MojoFailureException
+    {
+        ProjectBuildingRequest buildingRequest = buildBuildingRequest();
+        DefaultDependableCoordinate coordinate = getCoordinate();
+
+        try
+        {
+            getLog().info( "Retrieving classes of dependencies for " + artifact );
```
elharo, line 166: Not a helpful log message. In general we have way too much logging. Only log things the end user actually wants to see.

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,406 @@
...
+        {
+            throw new MojoExecutionException( "Couldn't download artifact: " + e.getMessage(), e );
+        }
+    }
+
+    public void printClassesFromArtifactResult( ArtifactResult result )
+            throws IOException
+    {
+        JarFile jarFile = new JarFile( result.getArtifact().getFile() );
+        Enumeration e = jarFile.entries();
```
elharo, line 195: e --> entries 

In general, avoid single letter variable names except for loop counters

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,406 @@
...
+
+        while ( e.hasMoreElements() )
+        {
+            JarEntry entry = (JarEntry) e.nextElement();
+            String name = entry.getName();
+
+            // filter out files that do not end in .class
+            if ( name.length() <= 6 || !name.substring( name.length() - 6 ).equals( ".class" ) )
```
elharo, line 203: name.endsWith(".class")

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
elharo: I think logging is how we usually do this. The log can be injected or tested.

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,406 @@
...
+                printClassesFromArtifactResult( result );
+            }
+        }
+        catch ( ArtifactResolverException | DependencyResolverException | IOException e )
+        {
+            throw new MojoExecutionException( "Couldn't download artifact: " + e.getMessage(), e );
+        }
+    }
+
+    public void printClassesFromArtifactResult( ArtifactResult result )
```
elharo, line 191: why is this public? Methods should have the most restrictive access possible.

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,406 @@
...
+            }
+
+            // remove .class from the end and change format to use periods instead of forward slashes
+            name = name.substring( 0, name.length() - 6 ).replace( '/', '.' );
+            getLog().info( name );
+        }
+        jarFile.close();
+    }
+
+    public ProjectBuildingRequest buildBuildingRequest()
```
elharo, line 215: why is this public? Methods should have the most restrictive access possible.

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,406 @@
...
+import org.apache.maven.shared.transfer.artifact.ArtifactCoordinate;
+import org.apache.maven.shared.transfer.artifact.DefaultArtifactCoordinate;
+import org.apache.maven.shared.transfer.artifact.resolve.ArtifactResolver;
+import org.apache.maven.shared.transfer.artifact.resolve.ArtifactResolverException;
+import org.apache.maven.shared.transfer.artifact.resolve.ArtifactResult;
+import org.apache.maven.shared.transfer.dependencies.DefaultDependableCoordinate;
+import org.apache.maven.shared.transfer.dependencies.DependableCoordinate;
+import org.apache.maven.shared.transfer.dependencies.resolve.DependencyResolver;
+import org.apache.maven.shared.transfer.dependencies.resolve.DependencyResolverException;
+import org.codehaus.plexus.util.StringUtils;
```
elharo, line 58: please don't use this class any more. apache commons stringutils if you must.

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,406 @@
...
+        return artifactCoordinate;
+    }
+
+    ArtifactRepository parseRepository( String repo, ArtifactRepositoryPolicy policy )
+            throws MojoFailureException
+    {
+        // if it's a simple url
+        String id = "temp";
+        ArtifactRepositoryLayout layout = getLayout( "default" );
+        String url = repo;
```
elharo, line 296: why copy this value?

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,406 @@
...
+        return dependencyResolver;
+    }
+
+    protected ArtifactResolver getArtifactResolver()
+    {
+        return artifactResolver;
+    }
+
+    /**
+     * @param artifact The artifact
```
elharo, line 360: doc comments begin with lower case, per Oracle guidelines

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
```diff
@@ -0,0 +1,406 @@
...
+    /**
+     * @param groupId The groupId.
+     */
+    public void setGroupId( String groupId )
+    {
+        this.coordinate.setGroupId( groupId );
+    }
+
+    /**
+     * @param artifactId The artifactId.
```
elharo, line 376: artifact ID

src/test/java/org/apache/maven/plugins/dependency/TestGetClassesMojo.java
```diff
@@ -0,0 +1,86 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import org.apache.maven.execution.MavenSession;
+import org.apache.maven.plugin.LegacySupport;
+import org.apache.maven.plugin.testing.stubs.MavenProjectStub;
+import org.apache.maven.settings.Server;
+import org.apache.maven.settings.Settings;
+import org.sonatype.aether.impl.internal.SimpleLocalRepositoryManager;
```
elharo, line 27: Tricky but can we replace this with org.eclipse equivalents? Probably not a 1:1 replacement.

src/test/java/org/apache/maven/plugins/dependency/TestGetClassesMojo.java
```diff
@@ -0,0 +1,86 @@
...
+import org.sonatype.aether.util.DefaultRepositorySystemSession;
+
+import java.io.File;
+
+public class TestGetClassesMojo
+        extends AbstractDependencyMojoTestCase
+{
+    GetClassesMojo mojo;
```
elharo, line 35: private

src/test/java/org/apache/maven/plugins/dependency/TestGetClassesMojo.java
```diff
@@ -0,0 +1,86 @@
...
+
+    protected void setUp()
+            throws Exception
+    {
+        // required for mojo lookups to work
+        super.setUp( "markers", false );
+
+        File testPom = new File( getBasedir(), "target/test-classes/unit/get-test/plugin-config.xml" );
+        assert testPom.exists();
```
elharo, line 44: NEVER use assert

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
bimargulies-google: It will work for a .war or any other 'renamed jar'. But the substitution of dots and slashes should perhaps be removed for non-.jar.

src/test/java/org/apache/maven/plugins/dependency/TestGetClassesMojo.java
bimargulies-google: Do you have an example to point at here?

src/test/java/org/apache/maven/plugins/dependency/TestGetClassesMojo.java
bimargulies-google: oh, whoops, sorry I missed this.

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
jonvolfson: I'm not sure if this is a useful feature to have but in case it is, I've set the default to false for now instead of removing it altogether.

src/test/java/org/apache/maven/plugins/dependency/TestGetClassesMojo.java
jonvolfson: private is not allowed for this test class

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
jonvolfson: fixed

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
jonvolfson: fixed

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
jonvolfson: fixed

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
jonvolfson: fixed

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
jonvolfson: More clarification? This is taken from the GetMojo.

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
jonvolfson: fixed

src/test/java/org/apache/maven/plugins/dependency/TestGetClassesMojo.java
jonvolfson: fixed

src/main/java/org/apache/maven/plugins/dependency/ListClassesMojo.java
```diff
@@ -0,0 +1,402 @@
...
+            String name = entry.getName();
+
+            // filter out files that do not end in .class
+            if ( !name.endsWith( ".class" ) )
+            {
+                continue;
+            }
+
+            // remove .class from the end and change format to use periods instead of forward slashes
+            name = name.substring( 0, name.length() - 6 ).replace( '/', '.' );
```
elharo, line 206: try to avoid reusing local variables. Here you have two things: an entry name and a class name, so two different variables make this clearer

src/main/java/org/apache/maven/plugins/dependency/ListClassesMojo.java
```diff
@@ -0,0 +1,402 @@
...
+        return dependencyResolver;
+    }
+
+    protected ArtifactResolver getArtifactResolver()
+    {
+        return artifactResolver;
+    }
+
+    /**
+     * @param artifact The artifact.
```
elharo, line 356: https://www.oracle.com/technical-resources/articles/java/javadoc-tool.html#tag

src/main/java/org/apache/maven/plugins/dependency/ListClassesMojo.java
```diff
@@ -0,0 +1,402 @@
...
+    @Component( role = ArtifactRepositoryLayout.class )
+    private Map<String, ArtifactRepositoryLayout> repositoryLayouts;
+
+    /**
+     * The repository system.
+     */
+    @Component
+    private RepositorySystem repositorySystem;
+
+    private DefaultDependableCoordinate coordinate = new DefaultDependableCoordinate();
```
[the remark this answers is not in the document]
jonvolfson: I do not as well. This was taken from the GetMojo. It seems this class is needed in order to properly use the artifact/dependency resolver when retrieving the specified artifact as DefaultArtifactCoordinate creates new problems.

src/main/java/org/apache/maven/plugins/dependency/ListClassesMojo.java
jonvolfson: fixed

src/main/java/org/apache/maven/plugins/dependency/ListClassesMojo.java
jonvolfson: fixed

src/main/java/org/apache/maven/plugins/dependency/ListClassesMojo.java
```diff
@@ -0,0 +1,402 @@
...
+            }
+
+            // remove .class from the end and change format to use periods instead of forward slashes
+            name = name.substring( 0, name.length() - 6 ).replace( '/', '.' );
+            getLog().info( name );
+        }
+        jarFile.close();
+    }
+
+    private ProjectBuildingRequest buildBuildingRequest()
```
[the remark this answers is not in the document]
jonvolfson: fixed

src/main/java/org/apache/maven/plugins/dependency/ListClassesMojo.java
```diff
@@ -0,0 +1,402 @@
...
+            {
+                ArtifactResult result = getArtifactResolver()
+                        .resolveArtifact( buildingRequest, toArtifactCoordinate( coordinate ) );
+
+                printClassesFromArtifactResult( result );
+            }
+        }
+        catch ( ArtifactResolverException | DependencyResolverException | IOException e )
+        {
+            throw new MojoExecutionException( "Couldn't download artifact: " + e.getMessage(), e );
```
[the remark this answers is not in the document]
jonvolfson: fixed

src/main/java/org/apache/maven/plugins/dependency/ListClassesMojo.java
```diff
@@ -0,0 +1,402 @@
...
+     *
+     */
+    @Parameter( property = "mdep.skip", defaultValue = "false" )
+    private boolean skip;
+
+    @Override
+    public void execute() throws MojoExecutionException, MojoFailureException
+    {
+        ProjectBuildingRequest buildingRequest = buildBuildingRequest();
+        DefaultDependableCoordinate coordinate = getCoordinate();
```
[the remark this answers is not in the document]
jonvolfson: fixed

src/main/java/org/apache/maven/plugins/dependency/ListClassesMojo.java
```diff
@@ -0,0 +1,402 @@
...
+
+        try
+        {
+            if ( isTransitive() )
+            {
+                Iterable<ArtifactResult> artifacts = getDependencyResolver()
```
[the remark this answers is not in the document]
jonvolfson: fixed

src/main/java/org/apache/maven/plugins/dependency/ListClassesMojo.java
```diff
@@ -0,0 +1,402 @@
...
+    {
+        return coordinate;
+    }
+
+    protected DependencyResolver getDependencyResolver()
+    {
+        return dependencyResolver;
+    }
+
+    protected ArtifactResolver getArtifactResolver()
```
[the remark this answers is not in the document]
jonvolfson: fixed, removed this method and other getters that weren't needed

src/main/java/org/apache/maven/plugins/dependency/ListClassesMojo.java
```diff
@@ -0,0 +1,402 @@
...
+    private String classifier;
+
+    /**
+     * The packaging of the artifact to download. Ignored if {@link #artifact} is used.
+     */
+    @Parameter( property = "packaging", defaultValue = "jar" )
+    private String packaging = "jar";
+
+    /**
+     * Repositories in the format id::[layout]::url or just URLs, separated by comma. ie.
```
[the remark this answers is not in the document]
jonvolfson: fixed

src/main/java/org/apache/maven/plugins/dependency/ListClassesMojo.java
```diff
@@ -0,0 +1,391 @@
...
+
+    /**
+     * Download transitively, retrieving the specified artifact and all of its dependencies.
+     */
+    @Parameter( property = "transitive", defaultValue = "false" )
+    private boolean transitive = false;
+
+    /**
+     * Skip plugin execution completely.
+     *
```
elharo, line 151: nit: remove blank line

src/main/java/org/apache/maven/plugins/dependency/ListClassesMojo.java
```diff
@@ -0,0 +1,391 @@
...
+            {
+                ArtifactResult result = artifactResolver
+                        .resolveArtifact( buildingRequest, toArtifactCoordinate( coordinate ) );
+
+                printClassesFromArtifactResult( result );
+            }
+        }
+        catch ( ArtifactResolverException | DependencyResolverException | IOException e )
+        {
+            throw new MojoExecutionException( "Couldn't download artifact " + artifact + ": " + e.getMessage(), e );
```
elharo, line 183: This message might not be quite accurate when you're downloading all transitive dependencies. That is, it could be one of the transitive deps we can't download instead of the original.

src/main/java/org/apache/maven/plugins/dependency/ListClassesMojo.java
```diff
@@ -0,0 +1,391 @@
...
+            if ( !entryName.endsWith( ".class" ) )
+            {
+                continue;
+            }
+
+            // remove .class from the end and change format to use periods instead of forward slashes
+            String className = entryName.substring( 0, entryName.length() - 6 ).replace( '/', '.' );
+            getLog().info( className );
+        }
+        jarFile.close();
```
elharo, line 208: use try-with-resources to guarantee that this is closed when an exception is thrown

src/main/java/org/apache/maven/plugins/dependency/ListClassesMojo.java
```diff
@@ -0,0 +1,391 @@
...
+        if ( layout == null )
+        {
+            throw new MojoFailureException( id, "Invalid repository layout", "Invalid repository layout: " + id );
+        }
+
+        return layout;
+    }
+
+    /**
+     * @return {@link #skip}
```
elharo, line 327: I don't think you need these getter methods, and probably not the setters

src/test/java/org/apache/maven/plugins/dependency/TestGetClassesMojo.java
```diff
@@ -0,0 +1,87 @@
...
+import org.junit.Assert;
+import org.sonatype.aether.impl.internal.SimpleLocalRepositoryManager;
+import org.sonatype.aether.util.DefaultRepositorySystemSession;
+
+import java.io.File;
+
+public class TestGetClassesMojo
+        extends AbstractDependencyMojoTestCase
+{
+    ListClassesMojo mojo;
```
elharo, line 36: make field private

src/main/java/org/apache/maven/plugins/dependency/ListClassesMojo.java
jonvolfson: I believe the original message was correct before I changed it to include artifact. If a transitive dependency wasn't able to be resolved, e.getMessage() should have included the correct one.

src/main/java/org/apache/maven/plugins/dependency/ListClassesMojo.java
jonvolfson: fixed

src/main/java/org/apache/maven/plugins/dependency/ListClassesMojo.java
jonvolfson: fixed

src/main/java/org/apache/maven/plugins/dependency/ListClassesMojo.java
jonvolfson: fixed

src/test/java/org/apache/maven/plugins/dependency/TestGetClassesMojo.java
jonvolfson: fixed

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
jonvolfson: fixed, forgot to comment

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
jonvolfson: fixed, forgot to comment

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
jonvolfson: fixed, forgot to comment

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
jonvolfson: fixed, forgot to comment

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
jonvolfson: fixed, forgot to comment

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
jonvolfson: fixed

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
jonvolfson: fixed

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
jonvolfson: fixed, removed unneeded variable

src/main/java/org/apache/maven/plugins/dependency/GetClassesMojo.java
jonvolfson: fixed

jira-importer:
Resolve #1075

PR #61, opened 2020-06-30
remove todos and other IDE inserted boilerplate

elharo:
@ian-lavallee


src/test/java/org/apache/maven/plugins/dependency/AbstractDependencyMojoTestCase.java
```diff
@@ -69,16 +64,11 @@ protected void tearDown()
...
-                // TODO Auto-generated catch block
                 e.printStackTrace();
                 fail( "Trying to remove directory:" + testDir + System.lineSeparator() + e.toString() );
             }
             assertFalse( testDir.exists() );
-
-            testDir = null;
         }
-
-        stubFactory = null;
```
ian-lavallee, line 81: 1. Why did we ever need to mark the stubFactory null, isn't garbage collection smart enough to deal with an object like this?
2. If not, why do we not need to manually mark stubFactory as null anymore?

PR #80, opened 2020-07-10
[MDEP-708] dependency:analyze should recommend test scope where possible

ian-lavallee:
This can't be merged until we have a new release and final version of maven-dependency-analyzer. For now it relies on a snapshot version.


ian-lavallee:
@elharo

ian-lavallee:
The test relies on the line count of the HTML file but I'm going to change that to verify the contents and just look for Junit in the NonTestScopedTestDependency section. Just wanted to get the PR in for you to see.

pom.xml
```diff
@@ -222,7 +222,8 @@ under the License.
     <dependency>
       <groupId>org.apache.maven.shared</groupId>
       <artifactId>maven-dependency-analyzer</artifactId>
-      <version>1.11.1</version>
+      <!-- NEED TO WAIT FOR RELEASE -->
+      <version>1.11.2-SNAPSHOT</version>
```
elharo, line 226: I've pushed this version to the snapshots repo. I'm checking on whether we can commit this based on that. I see at least one other plugin repo depending on a snapshot version.

elharo: given that we can depend on snapshot versions, are you comfortable with this PR being merged now?

ian-lavallee: Yes but I'm hoping to get @bimargulies  to review it first so he has more of my work to base his feedback off

pom.xml
```diff
@@ -427,14 +428,18 @@ under the License.
                 <pomExclude>tree-verbose/pom.xml</pomExclude>
               </pomExcludes>
               <pomIncludes>
+<!--
```
bimargulies-google, line 431: Delete it if we no longer want it.

src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
```diff
@@ -357,6 +359,15 @@ private boolean checkDependencies()
             warning = true;
         }
 
+        if ( !testArtifactsWithNonTestScope.isEmpty() )
+        {
+            getLog().warn( "Non-test scoped test only dependencies found:" );
```
bimargulies-google, line 364: Is this the message for 'have a broader scope' in the comment above? If so, perhaps change the comment to read more specifically that you are detecting "non-test scoped test-only dependencies'.

pom.xml
ian-lavallee: @elharo The PR can be merged now

src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
```diff
@@ -49,7 +49,7 @@
 
 /**
  * Analyzes the dependencies of this project and determines which are: used and declared; used and undeclared; unused
- * and declared. Also determines which dependencies have a broader scope than needed.
+ * and declared; non-test scoped test-only dependencies.
```
elharo, line 52: non-test scoped test-only dependencies --> compile scoped but only used in tests

ian-lavallee: It also works if a test only dependency is scoped with runtime, provided, or compile so how about; runtime, provided, or compile scoped but only used in tests?

elharo: That sounds like an issue. We probably don't want to suggest changing runtime scope to test scope since the analyzer can't detect how a runtime dependency is used.

ian-lavallee: Ok, the test fails if it's scoped with runtime but used in non test classes but it seems like it may be failing for the wrong reason. Gonna look into this more. If it needs to ignore runtime scopes that shouldn't be too complicated but will need to be changed in the actual dependency-analyzer project.

ian-lavallee:
Not because of the last commit relying on the new dependency-analyzer we need 1.11.3-SNAPSHOT for the analyzer now.

elharo:
FYI, I'm thinking the next release of maven-dependency-analyzer  perhaps should be 1.12.0.

ian-lavallee:
Should I change it to depend on 1.12.0 even though that's not released now then?

elharo:
No, just be aware

ian-lavallee:
This PR needs a 1.11.3-SNAPSHOT dependency-analyzer so that the test for compileScopedTestDependencies passes. Currently it fails because in 1.11.2 the dependency-analyzer considers runtime dependencies for testDependency analysis when it shouldn't

ian-lavallee:
@elharo can you trigger a CI build for this branch?

PR #95, opened 2020-08-14
[MDEP-644] Add tgf, dot, graphml support for verboseGraphSerializer

ian-lavallee:
Added DOT, TGF, and GraphML support for the verbose serializer


elharo:
Please change the title so it differs from #92

src/main/java/org/apache/maven/plugins/dependency/tree/VerboseGraphSerializer.java
```diff
@@ -42,30 +42,370 @@
...
     {
         Set<String> coordinateStrings = new HashSet<>();
         Map<String, String> coordinateVersionMap = new HashMap<>();
-        StringBuilder builder = new StringBuilder();
 
         // Use BFS to mirror how Maven resolves dependencies and use DFS to print the tree easily
         Map<DependencyNode, String> nodeErrors = getNodeConflictMessagesBfs( root, coordinateStrings,
                 coordinateVersionMap );
 
+        if ( "graphml".equals( outputType ) )
```
elharo, line 58: This long chain of if's makes me wonder if what we really need is a base abstract Serializer type and subclasses for each type.

src/main/java/org/apache/maven/plugins/dependency/tree/VerboseGraphSerializer.java
```diff
@@ -42,30 +42,370 @@
...
+                        + "<graphml xmlns=\"http://graphml.graphdrawing.org/xmlns\" "
+                + "xmlns:xsi=\"http://www.w3.org/2001/XMLSchema-instance\" "
+                + "xmlns:y=\"http://www.yworks.com/xml/graphml\" "
+                + "xsi:schemaLocation=\"http://graphml.graphdrawing.org/xmlns "
+                + "http://graphml.graphdrawing.org/xmlns/1.0/graphml.xsd\">"
+                + System.lineSeparator()
+                + "  <key for=\"node\" id=\"d0\" yfiles.type=\"nodegraphics\"/>"
+                + System.lineSeparator()
+                + "  <key for=\"edge\" id=\"d1\" yfiles.type=\"edgegraphics\"/>"
+                + System.lineSeparator()
```
elharo, line 91: to make this reproducible use a standard line separator, probably \n. No reason to use System line separator when writing XML in 2020.

src/main/java/org/apache/maven/plugins/dependency/tree/VerboseGraphSerializer.java
```diff
@@ -42,30 +42,370 @@
...
+                        result.append( " - " );
+                    }
+                    result.append( nodeErrors.get( child ) );
+                }
+                else if ( !visitedNodes.contains( child ) )
+                {
+                    visitedNodes.add( child );
+                    queue.add( child );
+                }
+                result.append( "\" ;" ).append( System.lineSeparator() );
```
elharo, line 275: I'm not familiar with the DOT format. What line separator does it want? use that, not a platform dependent separator.

src/main/java/org/apache/maven/plugins/dependency/tree/VerboseGraphSerializer.java
```diff
@@ -42,30 +42,370 @@
...
+    private void serializeTgfDfs( DependencyNode node, Map<DependencyNode, String> nodeErrors, StringBuilder nodes,
+                                  StringBuilder edges )
+    {
+        nodes.append( node.hashCode() ).append( " " );
+        String coordString = "";
+        boolean messageAdded = false;
+
+        if ( node.getArtifact().getProperties().containsKey( PRE_MANAGED_VERSION ) )
+        {
+            coordString = coordString.concat(
```
elharo, line 313: + or += might be clearer than concat. Up to you.

src/main/java/org/apache/maven/plugins/dependency/tree/VerboseGraphSerializer.java
```diff
@@ -42,30 +42,370 @@
...
+            }
+            else
+            {
+                nodes.append( coordString ).append( " - " ).append( nodeErrors.get( node ) );
+            }
+            nodes.append( ")" ).append( System.lineSeparator() );
+        }
+        else
+        {
+            nodes.append( coordString ).append( System.lineSeparator() );
```
elharo, line 362: again, pick a standard reproducible character

src/main/java/org/apache/maven/plugins/dependency/tree/VerboseGraphSerializer.java
```diff
@@ -42,30 +42,370 @@
...
+
+    private String serializeText( DependencyNode root, Map<DependencyNode, String> nodeErrors )
+    {
+        StringBuilder builder = new StringBuilder();
+
+        // deal with root first
+        Artifact rootArtifact = root.getArtifact();
+        builder.append( rootArtifact.getGroupId() ).append( ":" ).append( rootArtifact.getArtifactId() ).append(
+                ":" ).append( rootArtifact.getExtension() ).append( ":" ).append( rootArtifact.getVersion() ).append(
+                System.lineSeparator() );
```
elharo, line 398: \n

src/main/java/org/apache/maven/plugins/dependency/tree/VerboseGraphSerializer.java
```diff
@@ -42,30 +42,370 @@
             MANAGED_SCOPE = "managedScope";
 
     public String serialize( DependencyNode root )
+    {
+        return serialize( root, "" );
+    }
+
+    public String serialize( DependencyNode root, String outputType )
```
bimargulies-google, line 49: I don't suppose you'd consider an enum?

src/test/java/org/apache/maven/plugins/dependency/tree/VerboseGraphSerializerTest.java
```diff
@@ -250,4 +256,390 @@ public void testTreeWithVersionConflict() throws IOException
...
+        l3right.setChildren( Collections.singletonList( l4right ) );
+        l4right.setChildren( Collections.singletonList( l5right ) );
+        l5right.setChildren( Collections.singletonList( l6right ) );
+
+        String actual = serializer.serialize( root, "tgf" );
+        String expected = root.hashCode() + " org.example:root:jar:3.1.1" + System.lineSeparator()
+                + l1left.hashCode() + " org.duplicate:duplicate:xml:2:compile" + System.lineSeparator()
+                + l2left.hashCode() + " (org.duplicate:duplicate:xml:2:compile - omitted for duplicate)"
+                + System.lineSeparator() + l2middleLeft.hashCode()
+                + " (org.cycle:cycle:zip:3:compile - omitted for cycle)" + System.lineSeparator()
```
elharo, line 462: platform dependent

src/main/java/org/apache/maven/plugins/dependency/tree/VerboseGraphSerializer.java
ian-lavallee: Just committed that

src/main/java/org/apache/maven/plugins/dependency/tree/VerboseDependencyGraphBuilder.java
```diff
@@ -153,8 +152,8 @@ private void applyDependencyManagementDfs( Map<String, org.apache.maven.model.De
     private static String getDependencyManagementCoordinate( org.apache.maven.model.Dependency dependency )
     {
         StringBuilder string = new StringBuilder();
```
elharo, line 154: string -> builder

src/main/java/org/apache/maven/plugins/dependency/tree/VerboseGraphGraphmlSerializer.java
```diff
@@ -0,0 +1,174 @@
...
+    {
+        Set<String> coordinateStrings = new HashSet<>();
+        Map<String, String> coordinateVersionMap = new HashMap<>();
+        Map<DependencyNode, String> nodeErrors = getNodeConflictMessagesBfs( root, coordinateStrings,
+                coordinateVersionMap );
+
+        Set<DependencyNode> visitedNodes = new HashSet<>();
+        Queue<DependencyNode> queue = new LinkedList<>();
+        queue.add( root );
+        StringBuilder result = new StringBuilder( "<?xml version=\"1.0\" encoding=\"UTF-8\"?> "
```
elharo, line 47: you could use single quotes here instead of double quotes and make it a little easier to read by avoiding backslash escapes

elharo, line 47: are you sure the output encoding will be UTF-8?

ian-lavallee: I copied this based on the output from the non verbose graphml tree so I think we should keep it the same for now.

PR #114, opened 2020-10-18
Revert aborted release

elharo:
@michael-o Revert aborted 3.1.3 release. Next release should be 3.2.0.


michael-o (changes requested):
This doesn't make sense. 3.1.3 is already tagged. I see no need to abort. Just rerun the release process and select 3.2.0 and release and 3.2.1-SNAPSHOT as next dev version.

PR #135, opened 2021-06-01
[MDEP-568] go offline filter

bmarwell:

 - [X] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MDEP) filed 
       pulling in other changes.
 - [X] Format the pull request title like `[MDEP-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.
 - [X] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will 
 - [X] You have run the integration tests successfully (`mvn -Prun-its clean verify`).

 - [X] I hereby declare this contribution to be licenced under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

 - [X] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

### DIFF
--- a/src/it/projects/mdep-568-go-offline-exclude/pom.xml
+++ b/src/it/projects/mdep-568-go-offline-exclude/pom.xml
@@ -1,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one
+  ~ or more contributor license agreements.  See the NOTICE file
+  ~ distributed with this work for additional information
+  ~ regarding copyright ownership.  The ASF licenses this file
+  ~ to you under the Apache License, Version 2.0 (the
+  ~ "License"); you may not use this file except in compliance
+  ~ with the License.  You may obtain a copy of the License at
+  ~
+  ~   http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing,
+  ~ software distributed under the License is distributed on an
+  ~ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  ~ KIND, either express or implied.  See the License for the
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+
+  <groupId>org.apache.maven.its.dependency</groupId>
+  <artifactId>test</artifactId>
+  <version>1.0-SNAPSHOT</version>
+
+  <name>Test</name>
+  <description>
+    Test dependency:go-offline
+  </description>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-project</artifactId>
+      <version>2.0.6</version>
+    </dependency>
+    <dependency>
+      <groupId>non.existant.group</groupId>
+      <artifactId>fake</artifactId>
+      <version>1.0.0</version>
+    </dependency>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>non.existant.artifact</artifactId>
+      <version>1.0.0</version>
+    </dependency>
+  </dependencies>
+
+</project>
--- a/src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
@@ -1,0 +1,50 @@
+package org.apache.maven.plugins.dependency.filters;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.HashSet;
+import java.util.Locale;
+import java.util.Set;
+
+
+abstract class AbstractDependencyFilter implements DependencyFilter
+{
+    protected Set<String> splitValues( String csvValueList )
+    {
+        final String[] values = csvValueList.split( "," );
+        Set<String> excludeScope = new HashSet<>();
+
+        for ( String value : values )
+        {
+            // value is expected to be a scope, classifier, etc.
+            // thus assuming an english word. Do not rely on Locale.getDefault()!
+            final String cleanScope = value.trim().toLowerCase( Locale.ENGLISH );
+
+            if ( cleanScope.isEmpty() )
+            {
+                continue;
+            }
+
+            excludeScope.add( cleanScope );
+        }
+
+        return excludeScope;
+    }
+}
--- a/src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
@@ -19,6 +19,8 @@ package org.apache.maven.plugins.dependency.filters;
  * under the License.
  */
 
+import org.apache.maven.model.Dependency;
+
 import java.util.HashSet;
 import java.util.Locale;
 import java.util.Set;
@@ -26,6 +28,82 @@ import java.util.Set;
 
 abstract class AbstractDependencyFilter implements DependencyFilter
 {
+
+    protected final String includeIds;
+    protected final String excludeIds;
+
+    AbstractDependencyFilter( String includeIds, String excludeIds )
+    {
+        this.includeIds = includeIds == null ? "" : includeIds;
+        this.excludeIds = excludeIds == null ? "" : excludeIds;
+    }
+
+
+    @Override
+    public Set<Dependency> filter( Set<Dependency> dependencies )
+    {
+        Set<Dependency> filtered = new HashSet<>( dependencies );
+
+        filtered = filterincludeIds( filtered );
+        filtered = filterexcludeIds( filtered );
+
+        return filtered;
+    }
+
+    private Set<Dependency> filterexcludeIds( Set<Dependency> dependencies )
+    {
+        if ( excludeIds.trim().isEmpty() )
+        {
+            return dependencies;
+        }
+
+        final Set<String> excludedIds = splitExcludeIds( excludeIds );
+
+        Set<Dependency> filtered = new HashSet<>( dependencies.size() );
+        for ( Dependency dependency : dependencies )
+        {
+            if ( excludedIds.contains( getContainsProperty( dependency ) ) )
+            {
+                continue;
+            }
+
+            filtered.add( dependency );
+        }
+
+        return filtered;
+    }
+
+    private Set<Dependency> filterincludeIds( Set<Dependency> dependencies )
+    {
+        if ( includeIds.trim().isEmpty() )
+        {
+            return dependencies;
+        }
+
+        Set<String> includedIds = splitIncludeIds( includeIds );
+
+        Set<Dependency> filtered = new HashSet<>( dependencies.size() );
+        for ( Dependency dependency : dependencies )
+        {
+            if ( includedIds.contains( getContainsProperty( dependency ) ) )
+            {
+                filtered.add( dependency );
+            }
+        }
+
+        return filtered;
+    }
+
+    protected Set<String> splitExcludeIds( String excludeIds )
+    {
+        return splitValues( excludeIds );
+    }
+
+    protected Set<String> splitIncludeIds( String includeIds )
+    {
+        return splitValues( includeIds );
+    }
+
     protected Set<String> splitValues( String csvValueList )
     {
         final String[] values = csvValueList.split( "," );
@@ -47,4 +125,6 @@ abstract class AbstractDependencyFilter implements DependencyFilter
 
         return excludeScope;
     }
+
+    protected abstract String getContainsProperty( Dependency dependency );
 }
--- a/src/main/java/org/apache/maven/plugins/dependency/filters/ArtifactIdFilter.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/filters/ArtifactIdFilter.java
@@ -1,0 +1,98 @@
+package org.apache.maven.plugins.dependency.filters;
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *  http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.HashSet;
+import java.util.Set;
+
+import org.apache.maven.model.Dependency;
+
+/**
+ * Filters dependencies by artifact ID.
+ */
+public class ArtifactIdFilter extends AbstractDependencyFilter
+{
+    private final String includeArtifactIds;
+
+    private final String excludeArtifactIds;
+
+    public ArtifactIdFilter( String includeArtifactIds, String excludeArtifactIds )
+    {
+        this.includeArtifactIds = includeArtifactIds == null ? "" : includeArtifactIds;
+        this.excludeArtifactIds = excludeArtifactIds == null ? "" : excludeArtifactIds;
+    }
+
+    @Override
+    public Set<Dependency> filter( Set<Dependency> dependencies )
+    {
+        Set<Dependency> filtered = new HashSet<>( dependencies );
+
+        filtered = filterIncludeArtifactIds( filtered );
+        filtered = filterExcludeArtifactIds( filtered );
+
+        return filtered;
+    }
+
+    private Set<Dependency> filterExcludeArtifactIds( Set<Dependency> dependencies )
+    {
+        if ( excludeArtifactIds.trim().isEmpty() )
+        {
+            return dependencies;
+        }
+
+        final Set<String> excludedArtifactIds = splitValues( excludeArtifactIds );
+
+        Set<Dependency> filtered = new HashSet<>( dependencies.size() );
+        for ( Dependency dependency : dependencies )
+        {
+            if ( excludedArtifactIds.contains( dependency.getArtifactId() ) )
+            {
+                continue;
+            }
+
+            filtered.add( dependency );
+        }
+
+        return filtered;
+    }
+
+    private Set<Dependency> filterIncludeArtifactIds( Set<Dependency> dependencies )
+    {
+        if ( includeArtifactIds.trim().isEmpty() )
+        {
+            return dependencies;
+        }
+
+        Set<String> includedArtifactIds = splitValues( includeArtifactIds );
+
+        Set<Dependency> filtered = new HashSet<>( dependencies.size() );
+        for ( Dependency dependency : dependencies )
+        {
+            if ( includedArtifactIds.contains( dependency.getArtifactId() ) )
+            {
+                filtered.add( dependency );
+            }
+        }
+
+        return filtered;
+    }
+
+
+}
--- a/src/main/java/org/apache/maven/plugins/dependency/filters/ArtifactIdFilter.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/filters/ArtifactIdFilter.java
@@ -19,9 +19,6 @@ package org.apache.maven.plugins.dependency.filters;
  * under the License.
  */
 
-import java.util.HashSet;
-import java.util.Set;
-
 import org.apache.maven.model.Dependency;
 
 /**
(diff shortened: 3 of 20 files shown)


src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
```diff
@@ -0,0 +1,59 @@
+package org.apache.maven.plugins.dependency.filters;
```
elharo, line 1: Is this introducing a new package? Probably not needed.

src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.HashSet;
+import java.util.LinkedHashSet;
+import java.util.Locale;
+import java.util.Set;
+
+import org.apache.maven.model.Dependency;
+
+abstract class AbstractDependencyFilter implements DependencyFilter
+{
+    public boolean isDependencyIncluded( Dependency dependency )
```
elharo, line 31: Use `@Override` where appropriate

src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
```diff
@@ -0,0 +1,59 @@
...
+    {
+        Set<Dependency> set = new LinkedHashSet<>();
+        set.add( dependency );
+
+        set = filter( set );
+        return set.contains( dependency );
```
elharo, line 37: It feels like this could be done more directly without creating a new set.

src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    protected Set<String> splitValues( String csvScopeList )
+    {
+        final String[] scopes = csvScopeList.split( "," );
+        Set<String> excludeScope = new HashSet<>();
+
+        for ( String scope : scopes )
+        {
+            final String cleanScope = scope.trim().toLowerCase( Locale.ROOT );
```
elharo, line 47: Interesting. What does Locale.ROOT imply for the lowercasing?

src/main/java/org/apache/maven/plugins/dependency/filters/ClassifierFilter.java
```diff
@@ -0,0 +1,98 @@
...
+ * under the License.
+ */
+
+import java.util.HashSet;
+import java.util.Set;
+
+import org.apache.maven.model.Dependency;
+
+/**
+ * Filters dependencies by Classifier.
```
elharo, line 28: classifier (lower case)

src/main/java/org/apache/maven/plugins/dependency/filters/FilterDependencies.java
```diff
@@ -0,0 +1,91 @@
...
+
+import java.util.Collection;
+import java.util.HashSet;
+import java.util.List;
+import java.util.Set;
+
+import org.apache.maven.model.Dependency;
+
+/**
+ * Applies null to any number of filters to a given collection of dependencies.
```
elharo, line 33: I can't quite parse this comment.

src/main/java/org/apache/maven/plugins/dependency/filters/FilterDependencies.java
```diff
@@ -0,0 +1,91 @@
...
+
+        for ( DependencyFilter filter : filters )
+        {
+            try
+            {
+                filtered = filter.filter( filtered );
+            }
+            catch ( NullPointerException e )
+            {
+                // TODO: log
```
elharo, line 66: what an NPE here imply? why not let it bubble up?

src/main/java/org/apache/maven/plugins/dependency/filters/GroupIdFilter.java
```diff
@@ -0,0 +1,98 @@
...
+ * under the License.
+ */
+
+import java.util.HashSet;
+import java.util.Set;
+
+import org.apache.maven.model.Dependency;
+
+/**
+ * Filters dependencies by GroupId.
```
elharo, line 28: group ID.

elharo (commented):
All the filters should have unit tests.

src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
bmarwell: It should have been `Locale.ENGLISH`. I don't want to use no locale (ie system default), because in that case the `toLowerCase()` might behave differently.

src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
bmarwell: Yup, because code is similar to artifact-filter's ArtifactFilter which is also in its own package. Also, more filters could be introduced later. Is there a downside of creating a package I do not know?

src/main/java/org/apache/maven/plugins/dependency/filters/FilterDependencies.java
bmarwell: copied from FilterArtifacts, will just remove it.

src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
elharo: It tends to force methods to be public for use by other subpackages that otherwise wouldn't have to be public.

src/main/java/org/apache/maven/plugins/dependency/filters/ArtifactIdFilter.java
```diff
@@ -0,0 +1,98 @@
...
+/**
+ * Filters dependencies by artifact ID.
+ */
+public class ArtifactIdFilter extends AbstractDependencyFilter
+{
+    private final String includeArtifactId;
+
+    private final String excludeArtifactId;
+
+    public ArtifactIdFilter( String includeArtifactId, String excludeArtifactId )
```
elharo, line 36: are these plural? since they have to be split? That is are these really includeArtifactIds and excludeArtifactIds? If so, it might make sense to pass a collection of each instead. 

What happens if the same artifact ID is in both lists?

Group IDs not considered here? 


In either case it could use a doc comment.

bmarwell: Another copy/paste from other parts of the mojo. We have the choice of consistency vs doing it right.

Those come from the mojo as string parameter, thus can't change it. Have to split it here. But will change to plural.

bmarwell: The documentation does not state what should happen if it is in both lists. In this case it will be excluded. That should be fine and is not surprising behaviour.

src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
```diff
@@ -0,0 +1,50 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+import java.util.HashSet;
+import java.util.Locale;
+import java.util.Set;
+
+
+abstract class AbstractDependencyFilter implements DependencyFilter
```
antoniolucasnobar, line 27: Hello @bmarwell ,
	Since you created this abstract class, maybe it could contain more than just `splitValues` method. The inherited classes are very similar IMO; This way, this abstract class could have almost all logic here, with the filter, include and exclude methods; It is just a suggestion to avoid code duplication.

Below an example of the abstract class and one of its child. Even the `filterexcludeIds` and `filterincludeIds` could be simplified because they share similar logic.

As ScopeFilter has specific treatment to split include and exclue ids, I created two functions `splitIncludeIds` and `splitExcludeIds`

```java
	abstract class AbstractDependencyFilter implements DependencyFilter
	{
	protected final String includeIds;

	protected final String excludeIds;

	public AbstractDependencyFilter(String includeIds, String excludeIds)
	{
		this.includeIds = includeIds == null ? "" : includeIds;
		this.excludeIds = excludeIds == null ? "" : excludeIds;
	}


	@Override
	public Set<Dependency> filter(Set<Dependency> dependencies)
	{
		Set<Dependency> filtered = new HashSet<>(dependencies);

		filtered = filterincludeIds(filtered);
		filtered = filterexcludeIds(filtered);

		return filtered;
	}

	private Set<Dependency> filterexcludeIds(Set<Dependency> dependencies)
	{
		if (excludeIds.trim().isEmpty()) {
			return dependencies;
		}

		final Set<String> excludedIds = splitExcludeIds(excludeIds);

		Set<Dependency> filtered = new HashSet<>(dependencies.size());
		for (Dependency dependency : dependencies) {
			if (excludedIds.contains(getContainsProperty(dependency))) {
				continue;
			}

			filtered.add(dependency);
		}

		return filtered;
	}


	private Set<Dependency> filterincludeIds(Set<Dependency> dependencies)
	{
		if (includeIds.trim().isEmpty()) {
			return dependencies;
		}

		Set<String> includedIds = splitIncludeIds(includeIds);

		Set<Dependency> filtered = new HashSet<>(dependencies.size());
		for (Dependency dependency : dependencies) {
			if (includedIds.contains(getContainsProperty(dependency))) {
				filtered.add(dependency);
			}
		}

		return filtered;
	}

	protected Set<String> splitExcludeIds(String includeIds) {
		return splitValues(includeIds);
	}

	protected Set<String> splitIncludeIds(String includeIds) {
		return splitValues(includeIds);
	}

	protected Set<String> splitValues(String excludeIds) {
		// ommited here
	}

	abstract protected String getContainsProperty(Dependency dependency);

}
```

```java
	public class GroupIdFilter extends AbstractDependencyFilter
	{

		public GroupIdFilter(String includeGroupIds, String excludeGroupIds)
		{
			super(includeGroupIds, excludeGroupIds);
		}

	    protected String getContainsProperty(Dependency dependency) {
			return dependency.getGroupId();
		}

	}
```

bmarwell: I was thinking of closing this PR. We would need the same logic for other goals as well, and this goal could be replaced with the combination of two other goals.

I think you could just create a PR to my branch of you're interested to keep it up.

antoniolucasnobar: Created the PR to your branch: https://github.com/apache/maven-dependency-plugin/pull/145

Should I change anything else? There is a apparently unfinished thread about package names in June 1st. 

Do I have to sign any documents?  Please let me know.

PS.: I am sorry, butI did not understand why you are thinking in close this PR. Which goals are you talking about? Should they be a part in this PR? I think this PR solves the go-offline problem, which is the subject of MDEP-568.

src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
antoniolucasnobar: @elharo is this change a requirement for this PR to be accepted? I agree with @bmarwell approach (to be future proof), but if it is mandatory, I think it is possible to change packages.

bmarwell:
@elharo  can you please look over this PR again?

src/it/projects/mdep-568-go-offline-exclude/pom.xml
```diff
@@ -0,0 +1,57 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one
```
elharo, line 3: Why all the tildes in the license?

src/it/projects/mdep-568-go-offline-exclude/pom.xml
```diff
@@ -0,0 +1,57 @@
...
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-project</artifactId>
+      <version>2.0.6</version>
+    </dependency>
+    <dependency>
+      <groupId>non.existant.group</groupId>
```
elharo, line 46: nit: existent with an e

src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
```diff
@@ -0,0 +1,130 @@
...
+    {
+        Set<Dependency> filtered = new HashSet<>( dependencies );
+
+        filtered = filterincludeIds( filtered );
+        filtered = filterexcludeIds( filtered );
+
+        return filtered;
+    }
+
+    private Set<Dependency> filterexcludeIds( Set<Dependency> dependencies )
```
elharo, line 53: filterExcludeIds

src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
```diff
@@ -0,0 +1,130 @@
...
+        {
+            return dependencies;
+        }
+
+        final Set<String> excludedIds = splitExcludeIds( excludeIds );
+
+        Set<Dependency> filtered = new HashSet<>( dependencies.size() );
+        for ( Dependency dependency : dependencies )
+        {
+            if ( excludedIds.contains( getContainsProperty( dependency ) ) )
```
elharo, line 65: This is a bit awkward. Might be worth adding a local variable, renaming, the method, and/or refactoring with a more complete method

src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
```diff
@@ -0,0 +1,130 @@
...
+            {
+                continue;
```
elharo, line 67: don't use continue; just change the if to if not

src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
```diff
@@ -0,0 +1,130 @@
...
+            }
+
+            filtered.add( dependency );
+        }
+
+        return filtered;
+    }
+
+    private Set<Dependency> filterincludeIds( Set<Dependency> dependencies )
```
elharo, line 76: filterIncludeIds

src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
```diff
@@ -0,0 +1,130 @@
...
+        {
+            return dependencies;
+        }
+
+        Set<String> includedIds = splitIncludeIds( includeIds );
+
+        Set<Dependency> filtered = new HashSet<>( dependencies.size() );
+        for ( Dependency dependency : dependencies )
+        {
+            if ( includedIds.contains( getContainsProperty( dependency ) ) )
```
elharo, line 88: ditto

src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
```diff
@@ -0,0 +1,130 @@
...
+    {
+        return splitValues( includeIds );
+    }
+
+    protected Set<String> splitValues( String csvValueList )
+    {
+        final String[] values = csvValueList.split( "," );
+        Set<String> excludeScope = new HashSet<>();
+
+        for ( String value : values )
```
elharo, line 112: why is any case conversion necessary? Shouldn't this match including case?

src/main/java/org/apache/maven/plugins/dependency/filters/FilterDependencies.java
```diff
@@ -0,0 +1,85 @@
...
+    public FilterDependencies( DependencyFilter... filters )
+    {
+        this.filters = unmodifiableList( asList( filters ) );
+    }
+
+    /**
+     * Filter the given dependencies using the filters from this instance.
+     *
+     * @param dependencies The {@link Dependency}s to filter.
+     * @return the remaining dependencies as unmodifiable set.
```
elharo, line 53: no period

src/main/java/org/apache/maven/plugins/dependency/filters/ScopeFilter.java
```diff
@@ -0,0 +1,122 @@
...
+                return new HashSet<>( asList(
+                        Artifact.SCOPE_RUNTIME,
+                        Artifact.SCOPE_COMPILE ) );
+            case Artifact.SCOPE_COMPILE:
+                return new HashSet<>( asList(
+                        Artifact.SCOPE_COMPILE,
+                        Artifact.SCOPE_PROVIDED,
+                        Artifact.SCOPE_SYSTEM ) );
+            case Artifact.SCOPE_TEST:
+                throw new IllegalArgumentException( "invalid exclude scope: test. Did you mean compile?" );
```
elharo, line 72: why is this invalid?

src/main/java/org/apache/maven/plugins/dependency/filters/ScopeFilter.java
```diff
@@ -0,0 +1,122 @@
...
+                return new HashSet<>( singletonList( Artifact.SCOPE_PROVIDED ) );
+            case Artifact.SCOPE_SYSTEM:
+                return new HashSet<>( singletonList( Artifact.SCOPE_SYSTEM ) );
+            default:
+                throw new IllegalArgumentException( "Invalid Scope: " + scope );
+        }
+    }
+
+    /**
+     * runtime include scope gives runtime and compile dependencies,
```
elharo, line 83: Use HTML list or perhaps complete sentences.

src/main/java/org/apache/maven/plugins/dependency/filters/ScopeFilter.java
```diff
@@ -0,0 +1,122 @@
...
+     * compile include scope gives compile, provided, and system dependencies,
+     * test include scope gives all dependencies (equivalent to default),
+     * provided include scope just gives provided dependencies,
```
elharo, line 86: remove just

src/main/java/org/apache/maven/plugins/dependency/filters/TypeFilter.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+import org.apache.maven.model.Dependency;
+
+/**
+ * Filters dependencies by type.
+ */
+public class TypeFilter extends AbstractDependencyFilter
+{
+    public TypeFilter( String includeTypes, String excludeTypes )
```
elharo, line 29: api doc

src/main/java/org/apache/maven/plugins/dependency/resolvers/GoOfflineMojo.java
```diff
@@ -207,4 +239,14 @@ protected ArtifactsFilter getMarkedArtifactFilter()
     {
         return null;
     }
+
+    /**
+     * Returns a read-only set of dependencies used for going offline.
+     * @return an immutable set of dependencies used for going offline.
```
elharo, line 245: blank line before return

src/test/java/org/apache/maven/plugins/dependency/filters/ArtifactIdFilterTest.java
```diff
@@ -0,0 +1,69 @@
...
+import org.junit.Test;
+
+public class ArtifactIdFilterTest extends AbstractArtifactFilterTest
+{
+
+    @Test
+    public void testArtifactFilterExclude()
+    {
+        // given
+        final Dependency comSunTools = COM_SUN_TOOLS_SYSTEM_JAR;
```
elharo, line 39: local variables seem redundant here

src/test/java/org/apache/maven/plugins/dependency/filters/ArtifactIdFilterTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+        // then:
+        assertTrue( filtered.isEmpty() );
+    }
+
+
+    @Test
+    public void testArtifactFilterInclude()
+    {
+        // given
```
elharo, line 53: local variables seem redundant here

src/test/java/org/apache/maven/plugins/dependency/filters/ClassifierFilterTest.java
```diff
@@ -0,0 +1,74 @@
...
+import org.junit.Test;
+
+public class ClassifierFilterTest extends AbstractArtifactFilterTest
+{
+
+    @Test
+    public void testArtifactFilterExclude()
+    {
+        // given
+        final Dependency comSunTools = COM_SUN_TOOLS_SYSTEM_JAR;
```
elharo, line 38: local variables seem redundant here

src/test/java/org/apache/maven/plugins/dependency/filters/ClassifierFilterTest.java
```diff
@@ -0,0 +1,74 @@
...
+        assertEquals( 1, filtered.size() );
+        assertTrue( filtered.contains( comSunTools ) );
+    }
+
+
+    @Test
+    public void testArtifactFilterInclude()
+    {
+        // given
+        final Dependency comSunTools = COM_SUN_TOOLS_SYSTEM_JAR;
```
elharo, line 58: local variables seem redundant here

src/test/java/org/apache/maven/plugins/dependency/filters/GroupIdFilterTest.java
```diff
@@ -0,0 +1,69 @@
...
+        // then:
+        assertTrue( filtered.isEmpty() );
+    }
+
+
+    @Test
+    public void testGroupFilterInclude()
+    {
+        // given
+        final Dependency comSunTools = COM_SUN_TOOLS_SYSTEM_JAR;
```
elharo, line 54: local variables seem redundant here

src/test/java/org/apache/maven/plugins/dependency/filters/ScopeFilterTest.java
```diff
@@ -0,0 +1,86 @@
...
+import org.junit.Test;
+
+public class ScopeFilterTest extends AbstractArtifactFilterTest
+{
+
+    @Test
+    public void testScopeFilterExclude()
+    {
+        // given
+        final Dependency comSunTools = COM_SUN_TOOLS_SYSTEM_JAR;
```
elharo, line 39: local variables seem redundant here

src/test/java/org/apache/maven/plugins/dependency/filters/TypeFilterTest.java
```diff
@@ -0,0 +1,71 @@
...
+import org.apache.maven.model.Dependency;
+import org.junit.Test;
+
+public class TypeFilterTest extends AbstractArtifactFilterTest
+{
+
+    @Test
+    public void testTypeFilterExclude()
+    {
+        // given
```
elharo, line 38: local variables seem redundant here

src/test/java/org/apache/maven/plugins/dependency/resolvers/TestGoOfflineMojo.java
```diff
@@ -0,0 +1,105 @@
...
+import org.apache.maven.execution.MavenSession;
+import org.apache.maven.model.Dependency;
+import org.apache.maven.plugins.dependency.AbstractDependencyMojoTestCase;
+import org.apache.maven.plugins.dependency.utils.DependencyStatusSets;
+import org.apache.maven.project.MavenProject;
+
+public class TestGoOfflineMojo extends AbstractDependencyMojoTestCase
+{
+
+    protected void setUp() throws Exception
```
elharo, line 43: Override

elharo (commented):
assorted comments; though there seems to be some question from others about the overall approach that you might want to address first

src/it/projects/mdep-568-go-offline-exclude/pom.xml
antoniolucasnobar: fixed in https://github.com/apache/maven-dependency-plugin/pull/258

src/it/projects/mdep-568-go-offline-exclude/pom.xml
antoniolucasnobar: removed in https://github.com/apache/maven-dependency-plugin/pull/258. It is a pull request to this branch.

src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
antoniolucasnobar: fixed in https://github.com/apache/maven-dependency-plugin/pull/258

src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
antoniolucasnobar: fixed in https://github.com/apache/maven-dependency-plugin/pull/258

It is now like this:

```java
            boolean shouldFilterDependency = ids.contains( getId( dependency ) ) == expectedCondition;
            if ( shouldFilterDependency )
            {
                filtered.add( dependency );
            }
```
`expectedCondition`is true for the includeIds case anda false for the exclude. I created one method to remove the code duplication.

src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
antoniolucasnobar: fixed in https://github.com/apache/maven-dependency-plugin/pull/258

src/test/java/org/apache/maven/plugins/dependency/filters/ClassifierFilterTest.java
antoniolucasnobar: fixed in https://github.com/apache/maven-dependency-plugin/pull/258

src/test/java/org/apache/maven/plugins/dependency/filters/GroupIdFilterTest.java
antoniolucasnobar: fixed in https://github.com/apache/maven-dependency-plugin/pull/258

src/test/java/org/apache/maven/plugins/dependency/filters/ScopeFilterTest.java
antoniolucasnobar: fixed in https://github.com/apache/maven-dependency-plugin/pull/258

src/test/java/org/apache/maven/plugins/dependency/filters/TypeFilterTest.java
antoniolucasnobar: fixed in https://github.com/apache/maven-dependency-plugin/pull/258

src/test/java/org/apache/maven/plugins/dependency/resolvers/TestGoOfflineMojo.java
antoniolucasnobar: fixed in https://github.com/apache/maven-dependency-plugin/pull/258

antoniolucasnobar:
@elharo I fixed in https://github.com/apache/maven-dependency-plugin/pull/258 most of your questions.
Also, I resolved the conflicts with master.

About the overall aproach, I was never able to fully understand what is the problem with this one. @bmarwell could you please help with this?

jshbrntt:
Is this going to be merged anytime soon?

It would be really helpful to install and cache dependencies prior to `mvn package`.

src/it/projects/mdep-568-go-offline-exclude/pom.xml
elharo: I don't understand. I still see them. Do you need to merge to master?

src/main/java/org/apache/maven/plugins/dependency/filters/AbstractDependencyFilter.java
elharo: Yes, packaghes are overused with bad consequences

elharo (changes requested):
I don't understand what's going on with this commit. Please make sure it's up to date against master, and comments are addressed here, not in other PRs.

bmarwell:
This is old. If interested, please do a new PR. I'll leave the branch for a while.

jira-importer:
Resolve #1004

PR #205, opened 2022-03-07
Exclude vulnerable commons-beanutils-1.7.0 from maven-reporting-impl …

gyb123456:
I saw team has remove the dependency of doxia-site-renderer and upgrade maven-reporting-impl from 3.0 to 3.1, but the 
commons-beanutils-1.7.0 is still in maven-reporting-impl:3.1. Hope we could exclude the vulnerable version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -164,6 +164,12 @@ under the License.
       <groupId>org.apache.maven.reporting</groupId>
       <artifactId>maven-reporting-impl</artifactId>
       <version>3.1.0</version>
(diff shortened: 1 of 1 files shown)


slawekjaranowski:
We have direct dependency to commons-beanutils 1.9.4, so excluding is not needed.

https://github.com/apache/maven-dependency-plugin/pull/177
https://issues.apache.org/jira/browse/MDEP-797

But thanks for proposition and collaborations.

PR #218, opened 2022-05-20
[MDEP-808] restrict dependency analysis by group

chadwick00:
Full write up of the use case on MDEP-808. Short version is that I've added a new {{includeDependencies}} parameter to allow the scope of the warnings from the analyze goal to be limited to the include dependencies.

       pulling in other changes.
       commit message.


src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
```diff
@@ -215,7 +216,7 @@
      * @since 2.10
      */
     @Parameter
-    private String[] ignoredUsedUndeclaredDependencies = new String[0];
+    private final String[] ignoredUsedUndeclaredDependencies = new String[0];
```
slawekjaranowski, line 219: Please don't use `final` for Mojo parameters. It can be inline by compiler ...

slawekjaranowski (changes requested):
Pleas add assertions for IT test

src/it/projects/analyze-include-dependency/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+  </description>
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-project</artifactId>
```
elharo, line 42: why so old?

src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
```diff
@@ -266,7 +267,30 @@ public abstract class AbstractAnalyzeMojo
...
+     * The filter syntax is:
+     *
+     * <pre>
+     * [groupId]:[artifactId]:[type]:[version]
+     * </pre>
+     *
+     * where each pattern segment is optional and supports full and partial <code>*</code> wildcards. An empty pattern
+     * segment is treated as an implicit wildcard. *
+     * <p>
+     * For example, <code>org.apache.*</code> will match all artifacts whose group id starts with
```
elharo, line 285: will match --> matches (tech writing lives in the eternal present)

src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
```diff
@@ -266,7 +267,30 @@ public abstract class AbstractAnalyzeMojo
...
+     * <code>org.apache.</code>, and <code>:maven-dependency-plugin</code> will result in the analysis being applied
```
elharo, line 286: This is confusing. Try using an example of an unrelated library instead like JodaTime.

src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
```diff
@@ -266,7 +267,30 @@ public abstract class AbstractAnalyzeMojo
     // defaultValue value on @Parameter - not work with Maven 3.2.5
     // When is set defaultValue always win, and there is no possibility to override by plugin configuration.
     @Parameter
-    private List<String> ignoredPackagings = Arrays.asList( "pom", "ear" );
+    private final List<String> ignoredPackagings = Arrays.asList( "pom", "ear" );
+
+
+    /**
+     * List of dependencies to be included. The result of this list is then applied to the ignore parameters.
```
elharo, line 274: This isn't clear that everything is included by default when this isn't set.

src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
```diff
@@ -438,6 +478,40 @@ private boolean checkDependencies()
             warning = true;
         }
 
+        // log dependencies that weren't included
+        if ( verbose && !notIncludedUsedDeclared.isEmpty() )
```
elharo, line 482: why are these logged? I thought the point of this PR was to avoid reporting on these.

src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
```diff
@@ -621,7 +695,15 @@ private void writeScriptableOutput( Set<Artifact> artifacts )
         }
     }
 
-    private List<Artifact> filterDependencies( Set<Artifact> artifacts, String[] excludes )
+
+    /**
+     * Filter for artifacts that don't match the <code>exclude</code> criteria.
+     *
+     * @param artifacts filtered by elements that don't match the <code>excludes</code> argument
```
elharo, line 702: exclude or excludes? You have both.

src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
```diff
@@ -638,4 +720,35 @@ private List<Artifact> filterDependencies( Set<Artifact> artifacts, String[] exc
...
+     *
+     * @param artifacts filtered for elements that do match the criteria
+     * @param includes the filter to be applied
+     * @return the list of artifacts that didn't match the criteria
+     */
+    private List<Artifact> filterDependencies( Set<Artifact> artifacts, String[] includes )
+    {
+        List<Artifact> result = new ArrayList<>();
+
+        if ( includes.length > 0 )
```
elharo, line 737: !isEmpty()

src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
```diff
@@ -638,4 +720,35 @@ private List<Artifact> filterDependencies( Set<Artifact> artifacts, String[] exc
...
+        {
+          ArtifactFilter filter = new StrictPatternIncludesArtifactFilter( Arrays.asList( includes ) );
+
+            for ( Iterator<Artifact> it = artifacts.iterator(); it.hasNext(); )
+            {
+                Artifact artifact = it.next();
+                if ( !filter.include( artifact ) )
+                {
+                    it.remove();
```
elharo, line 746: why the remove call? Doesn't seem needed.

elharo (changes requested):
What happens if the same artifact matches both includes and excludes? This needs to be clearly documented.

Also, this PR should include the necessary changes to the project docs, not just the code.

src/it/projects/analyze-include-dependency/pom.xml
chadwick00: I think it was just that I wanted to valid project and didn't particularly care about the versions within the project. I can search for a substitution variable to keep in current if having this always lag behind is an eye-sore come review time.

src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
chadwick00: Done.

chadwick00: Although looking back through the code again, I probably did this to remind myself that these aren't mutated by the code. Presumably only a mad-man would write code that would change the contents of a Mojo parameter, even if the compiler allowed them to.

src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
chadwick00: Yes, I've been musing on this point. The default behaviour prior to this change is that everything is included. That can't change without requiring a pom changes downstream. So this is strictly:
 
"_an override to the ignore_"

That's a bit mind-bending if you don't have the context. I think the best we can do is to provide a recipe in the Javadoc describing an example use case, such as the one we're going to use it for. Broadly, everything _ignored_, but this one Maven co-ordinate _in_.

It doesn't feel like if we were designing this anew we'd start from here, but without a brain-wave on how to prevent a backward incompatible change, this will be the approach.

chadwick00: Thinking through some of the use-cases further and wanting more fine-grained control, we're essentially building sub-graphs in the dependency graph. I think that is expressible using Maven co-ordinates, but would be good to back these thoughts with some tests.

src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
chadwick00: Not a collection. Happy to call Arrays.asList (includes) earlier if you prefer.

src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
chadwick00: The original `filterDependencies` method both mutates the `artifacts` argument and indicates what is has removed in the `result`. I didn't want to change the stateful nature of this methods because that would be a pretty serious overhaul to the code. So the remove call is needed to remove `artifacts` that don't match the include argument.

src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
chadwick00: These are logged in the verbose section only. In the same way that the [existing verbose section](https://github.com/apache/maven-dependency-plugin/blob/master/src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java#L403) for the explicitly ignored dependencies, I think you'd also want to understand what was been ignore implicitly by not being covered within the include block.

src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
chadwick00: Done by copying the example Javadoc above. Decided the purpose of this section of the Javadoc is to explain the Maven co-ordinate and wildcard combination rather than the specific behaviour of the parameter.

src/it/projects/analyze-include-dependency/pom.xml
chadwick00: I've tried to update the version of the all the dependencies @mavenVersion@, but at least one of the dependencies doesn't like it. I can try harder by choosing alternative dependencies, but I suppose the benefit of keeping these at an old version is that the referenced classes within dependency won't deprecated, as [has been done elsewhere](https://github.com/apache/maven-dependency-plugin/blob/master/src/it/projects/analyze-ignore-unused-declared-dependency/pom.xml#L43).

src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
chadwick00: Decided that building sub-graphs was way to complicated, so sticking with the original changes, which is consistent with the standard Maven filters.... [include then exclude](https://github.com/apache/maven-common-artifact-filters/blob/9d692d49b35014a2d308c78894ee502bccd0361e/src/main/java/org/apache/maven/shared/artifact/filter/collection/AbstractArtifactFeatureFilter.java#L63).

src/site/apt/examples/control-dependencies-under-dependency-analysis.apt.vm
```diff
@@ -37,6 +37,11 @@ Exclude dependencies from dependency analysis
   The plugin can then be configured to ignore dependencies that are
   "declared but unused", "undeclared but used", and "non-test scoped"
   in selected list or in all simultaneously.
+  
+  Another case is when mature projects have accumulated superfluous dependencies or
+  relied upon transitivity for their dependencies. Where fixing all the dependency 
+  problems up-front is not viable the plugin can be configured to only raise
```
elharo, line 43: comma after viable

src/site/apt/examples/control-dependencies-under-dependency-analysis.apt.vm
```diff
@@ -57,6 +62,11 @@ Exclude dependencies from dependency analysis
             </goals>
             <configuration>
               <failOnWarning>true</failOnWarning>
+              
+              <!-- limit warnings to dependencies for group com.google.code.findbugs -->
+              <includeDependencies>
+                <includeDependency>com.google.code.findbugs</includeDependency>
```
elharo, line 68: includeDependency --> groupId

src/site/apt/examples/control-dependencies-under-dependency-analysis.apt.vm
```diff
@@ -37,6 +37,11 @@ Exclude dependencies from dependency analysis
   The plugin can then be configured to ignore dependencies that are
   "declared but unused", "undeclared but used", and "non-test scoped"
   in selected list or in all simultaneously.
+  
+  Another case is when mature projects have accumulated superfluous dependencies or
+  relied upon transitivity for their dependencies. Where fixing all the dependency 
+  problems up-front is not viable the plugin can be configured to only raise
+  warnings for dependencies meeting "include" criteria.
```
elharo, line 44: This needs to be expanded. Based solely on this text — that is, without reading the code or the JIRA — I do not know which dependencies are and are not checked.

elharo (commented):
I'm still skeptical of this feature. What is the problem solved by not warning on all unused dependencies? Unused dependencies don't break the build. If someone can't fix them all right now, so what?

elharo:
Making a decision not to do this. Discussion in Jira

jira-importer:
Resolve #1317

PR #219, opened 2022-05-24
[MDEP-602] - Log dependency warnings as errors when failOnWarning is set

hisener:
As the ticket describes, having dependency warnings fail the build is somewhat unexpected to many developers. This change logs those warnings as errors when the `failOnWarning` parameter is set to true.

---

       pulling in other changes.
       commit message.

### DIFF
--- a/src/it/projects/mdep-602-log-on-error-level-when-failOnWarning/pom.xml
+++ b/src/it/projects/mdep-602-log-on-error-level-when-failOnWarning/pom.xml
@@ -1,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+Licensed to the Apache Software Foundation (ASF) under one
+or more contributor license agreements.  See the NOTICE file
+distributed with this work for additional information
+regarding copyright ownership.  The ASF licenses this file
+to you under the Apache License, Version 2.0 (the
+"License"); you may not use this file except in compliance
+with the License.  You may obtain a copy of the License at
+
+  http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing,
+software distributed under the License is distributed on an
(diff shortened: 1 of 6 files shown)


src/it/projects/mdep-602-log-on-error-level-when-failOnWarning/pom.xml
```diff
@@ -0,0 +1,55 @@
...
+
+  <properties>
+    <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
+  </properties>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.apache.maven</groupId>
+      <artifactId>maven-project</artifactId>
+      <version>2.0.6</version>
```
slawekjaranowski, line 39: please use something newer - 3.2.5

slawekjaranowski (changes requested):
Do we have a test with `failOnWarning = false` and assertions with warning log level?
Maybe in other test we can add assertions for warning.

src/it/projects/mdep-602-log-on-error-level-when-failOnWarning/pom.xml
hisener: I copy-pasted this from another test. :grimacing: I can bump the version, but 3.2.5 doesn't seem available on Maven Central. :eyes: https://search.maven.org/artifact/org.apache.maven/maven-project

slawekjaranowski: ok, leave as is. 

It will be different task to upgrade Maven version in tests 😄

hisener:
Added a commit for adding assertions, PTAL.

Shall I squash commits into one?

slawekjaranowski (approved):

slawekjaranowski:
@hisener thanks

PR #245, opened 2022-09-11
[MDEP-716] - TestListClassesMojo logs too much

raupachz:
Hello there,

attached is a PR for [MDEP-716](https://issues.apache.org/jira/browse/MDEP-716). This is an `up-for-grabs` issue. 

This code uses Mockito and creates a mock of the `org.apache.maven.plugin.logging.Log.java`. A redirect of `System.out` is not needed. This has also the nice benefit of argument matching. 

---

       pulling in other changes.
       commit message.


slawekjaranowski (commented):
Are files different?
 src/test/resources/unit/list-test/testListClassesNotTransitive.txt
 src/test/resources/unit/list-test/testListClassesTransitive.txt

If yes we have the same output for different inputs for `transitive`...

raupachz:
Hi @slawekjaranowski 

no, the files are the same. I did not check the output. Before the output was just written to `System.out` without any checking. I treated every unit-test as a separate unit with an individual input file. Made sense to me. Even if this results in duplicate file content.

If `org.apache.commons:commons-lang3:3.6` has no transitive dependencies then the output should be the same - to my understanding.

slawekjaranowski:
> If `org.apache.commons:commons-lang3:3.6` has no transitive dependencies then the output should be the same - to my understanding.

good understanding, 

So I see some improvement here
 - test for artifacts without transitive dependencies - as we have now, only test name / description should point such case
- test for artifact with transitive dependencies

of course it is job for next PR

jira-importer:
Resolve #1209

PR #255, opened 2022-10-20
[MDEP-832] - Remove commons-collections-4

khmarbaise:

       pulling in other changes.
       commit message.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -260,11 +260,6 @@ under the License.
       <artifactId>commons-collections4</artifactId>
       <version>4.4</version>
     </dependency>
-    <dependency>
-      <groupId>org.apache.commons</groupId>
-      <artifactId>commons-lang3</artifactId>
-      <version>3.12.0</version>
-    </dependency>
 
     <!-- dependencies to annotations -->
     <dependency>
--- a/src/main/java/org/apache/maven/plugins/dependency/analyze/AnalyzeDuplicateMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/analyze/AnalyzeDuplicateMojo.java
@@ -20,14 +20,14 @@ package org.apache.maven.plugins.dependency.analyze;
 
 import java.io.IOException;
 import java.io.Reader;
-import java.util.ArrayList;
 import java.util.Collections;
+import java.util.HashSet;
 import java.util.Iterator;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
+import java.util.stream.Collectors;
 
-import org.apache.commons.collections4.CollectionUtils;
 import org.apache.maven.model.Dependency;
 import org.apache.maven.model.Model;
 import org.apache.maven.model.io.xpp3.MavenXpp3Reader;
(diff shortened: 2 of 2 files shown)


src/main/java/org/apache/maven/plugins/dependency/analyze/AnalyzeDuplicateMojo.java
```diff
@@ -151,15 +150,9 @@ private void createMessage( Set<String> duplicateDependencies, StringBuilder sb,
...
-            modelDependencies2.add( dep.getManagementKey() );
-        }
-
-        // @formatter:off
+        List<String> modelDependencies2 =
+                modelDependencies.stream().map( Dependency::getManagementKey ).collect( Collectors.toList() );
         return new LinkedHashSet<>(
-                CollectionUtils.disjunction( modelDependencies2, new LinkedHashSet<>( modelDependencies2 ) ) );
-        // @formatter:on
+                Util.symmetricDifference( modelDependencies2, new LinkedHashSet<>( modelDependencies2 ) ) );
```
gnodet, line 156: I struggled on that one.
Given the two sets are the same, I think the result will always be an empty set.
So I would just simply remove that call.

khmarbaise: No they are not the same. The first list (modelDependencies2) contains two entries (junit) while the hashset contains only a single entry of that and the result must be the single junit-jar difference. Check the unit test: `TestAnalyzeDuplicateMojo` testcase `testDuplicate` and `testDuplicate2` which references `src/test/resources/unit/duplicate-dependencies/plugin-config.xml` and the `src/test/resources/unit/duplicate-dependencies/plugin-config2.xml` which contains exactly those cases.

gnodet: Ah, got it.  But then, wouldn't it be simpler and more understandable :
```
    /**
     * Remove exactly one instance of each element from the give list. 
     */
    static <O> List<O> symmetricDifference( Collection<O> elements )
    {
        List<O> list = new ArrayList<>( elements );
        Set<O> set = new HashSet<>( elements );
        list.removeIf( set::remove );
        return list;
    }
```

gnodet: It could even be inlined with a simple comment...

gnodet: The shortest form could be:
```
    private Set<String> findDuplicateDependencies( List<Dependency> modelDependencies )
    {
        List<String> modelDependencies2 = modelDependencies.stream()
                .map( Dependency::getManagementKey ).collect( Collectors.toList() );
        // remove one instance of each element from the list
        modelDependencies2.removeIf( new HashSet<>( modelDependencies2 )::remove );
       // keep a single instance of each duplicate
        return new LinkedHashSet<>( modelDependencies2 );
    }
```

bmarwell (changes requested):
I agree with Guillaume. Please simplify (or remove) the util class with the code he provided.
After that, consider my changes request obsolete. :)

pom.xml
```diff
@@ -273,6 +267,12 @@ under the License.
       <version>4.13.2</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
```
elharo, line 270: can we stick to pure JUnit 4.x only? Test library proliferation is simply confusing

src/main/java/org/apache/maven/plugins/dependency/analyze/Util.java
```diff
@@ -0,0 +1,58 @@
...
+import java.util.Map;
+import java.util.Set;
+
+import static java.util.function.Function.identity;
+import static java.util.stream.Collectors.collectingAndThen;
+import static java.util.stream.Collectors.groupingBy;
+import static java.util.stream.Collectors.summingInt;
+import static java.util.stream.Collectors.toList;
+
+class Util
```
elharo, line 34: This needs a more specific name. Arguably this can be a private method in the class that uses it.

pom.xml
```diff
@@ -260,11 +260,6 @@ under the License.
       <artifactId>commons-collections4</artifactId>
```
vtintillier, line 260: didn't you mean to remove this one?

elharo:
There were a number of pom changes in flight so it looks like some of the PRs got a little mixed up when resolving merge conflicts, but I think it is removed at HEAD.

jira-importer:
Resolve #1337

PR #268, opened 2022-11-27
[MDEP-831] remove unused beanutils dependency

elharo:
@khmarbaise

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -326,12 +326,6 @@ under the License.
       <version>${slf4j.version}</version>
       <scope>test</scope>
     </dependency>
-    <dependency>
-      <groupId>commons-beanutils</groupId>
-      <artifactId>commons-beanutils</artifactId>
-      <version>1.9.4</version>
-    </dependency>
-
   </dependencies>
 
   <build>


slachiewicz (approved):

elharo:
CI looks good: https://ci-maven.apache.org/job/Maven/job/maven-box/job/maven-dependency-plugin/job/bean/

slawekjaranowski:
@elharo , @slachiewicz 

It was added in order to override transitive version, now we have version 1.7.0 - please examine dependency tree

Why we need newer version ... because of CVE ...

Probably better place will be dependencyManagement for such case.

elharo:
Adding an extra dependency is not the right way to handle this. DependencyManagement might be better but is not really right for this case either. This needs to be fixed in whatever dependency is pulling in the old version.

elharo:
Seems like the correct way to handle this is by releasing org.apache.maven.doxia:doxia-site-renderer:2.0.0 and then upgrading the dependency plugin to that version.

slawekjaranowski:
You right fixing such issue at source is the best way, but until it happens we should have some workaround.

PR #290, opened 2023-03-05
use Apache Commons StringUtil

elharo:


elharo:
Failure might be flaky or perhaps platform dependent?

[INFO] -------------------------------------------------
Error:  The following builds failed:
Error:  *  analyze-testDependencyWithNonTestScope/pom.xml
[INFO] -------------------------------------------------

elharo:
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.10.1:compile (default-compile) on project nonTestScopeTestArtifactsAnalysis: Fatal error compiling: error: release version 1.7 not supported -> [Help 1]

elharo:
It does look like the failing CI is broken at head base don other failing PRs

michael-o (approved):

slawekjaranowski:
Maybe it is technically easy change .... but we should report issue for such, and each project should have own issue
Now I see a mess like in #299 - which is connected with issue from MNG

jira-importer:
Resolve #1353

PR #293, opened 2023-03-11
update IT analyze-testDependencyWithNonTestScope to Java 8 to fix broken CI

elharo:


elharo:
Hmm, still not fixed.

elharo:
actual fix got merged to master by accident since we don't protect master branches, but at least it worked.

PR #304, opened 2023-04-17
[MDEP-856] Upgrade transitive dependency to commons-text from 1.3 to 1.10.0

slachiewicz:


pom.xml
```diff
@@ -165,6 +165,11 @@ under the License.
       <artifactId>maven-reporting-impl</artifactId>
       <version>3.2.0</version>
     </dependency>
+    <dependency>
+      <groupId>org.apache.commons</groupId>
+      <artifactId>commons-text</artifactId>
+      <version>1.10.0</version>
+    </dependency>
```
slawekjaranowski, line 172: for transitive version override I like to use dependency Management, it is more clear that we only manage of version but it is not a direct needed dependency.

slachiewicz: thx - updated

slawekjaranowski (approved):

PR #325, opened 2023-05-22
[MDEP-799] tree: add optional output type json

MartinWitt:

 - [X] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MDEP) filed 
       pulling in other changes.

I will squash and update the commit in the end. There is still open discussion. 

 - [X] Format the pull request title like `[MDEP-XXX] - Fixes bug in ApproximateQuantiles`,
       commit message.
 - [X] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will 

 - [X] I hereby declare this contribution to be licensed under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

 - [X] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

I sent the mail for this a few minutes ago so it should be signed soon.

I tried to understand the way you write unit/integration tests but it looks a bit complicated. What is the correct way to add a testcase? Best case, I can add a pom and check if the model written as json -> read by any JSON parser is equal to the model before.

The JSON files currently exist for you to see the current printed format. It will be deleted afterward.

This PR adds JSON output type to the maven-depdenceny-plugin with the goal `tree`.

Supersedes #207

### DIFF
--- a/src/main/java/org/apache/maven/plugins/dependency/tree/JsonDependencyNodeVisitor.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/tree/JsonDependencyNodeVisitor.java
@@ -26,10 +26,18 @@ import org.apache.maven.artifact.Artifact;
 import org.apache.maven.shared.dependency.graph.DependencyNode;
 import org.apache.maven.shared.dependency.graph.traversal.DependencyNodeVisitor;
 
+/**
+ * A dependency node visitor that serializes visited nodes to a writer using the JSON format.
+ */
 public class JsonDependencyNodeVisitor extends AbstractSerializingVisitor implements DependencyNodeVisitor {
 
+    private static final String LINE_SEPARATOR = "\n";
     private String indentChar = " ";
 
+    /**
+     * Creates a new instance of {@link JsonDependencyNodeVisitor}. The writer will be used to write the output.
+     * @param writer  the writer to write to
+     */
     public JsonDependencyNodeVisitor(Writer writer) {
         super(writer);
     }
@@ -50,9 +58,9 @@ public class JsonDependencyNodeVisitor extends AbstractSerializingVisitor implem
     private void writeRootNode(DependencyNode node, Writer writer) {
         int indent = 2;
         StringBuilder sb = new StringBuilder();
-        sb.append("{").append(System.lineSeparator());
+        sb.append("{").append(LINE_SEPARATOR);
         writeNode(indent, node, sb);
-        sb.append("}").append(System.lineSeparator());
+        sb.append("}").append(LINE_SEPARATOR);
         try {
             writer.write(sb.toString());
         } catch (IOException e) {
@@ -79,21 +87,21 @@ public class JsonDependencyNodeVisitor extends AbstractSerializingVisitor implem
      * @param sb  the string builder to append to
      */
     private void writeChildren(int indent, DependencyNode node, StringBuilder sb) {
-        sb.append(indent(indent)).append("\"children\": [").append(System.lineSeparator());
+        sb.append(indent(indent)).append("\"children\": [").append(LINE_SEPARATOR);
         indent += 2;
         for (int i = 0; i < node.getChildren().size(); i++) {
             DependencyNode child = node.getChildren().get(i);
             sb.append(indent(indent));
-            sb.append("{").append(System.lineSeparator());
+            sb.append("{").append(LINE_SEPARATOR);
             writeNode(indent + 2, child, sb);
             sb.append(indent(indent)).append("}");
             // we skip the comma for the last child
             if (i != node.getChildren().size() - 1) {
                 sb.append(",");
             }
-            sb.append(System.lineSeparator());
+            sb.append(LINE_SEPARATOR);
         }
-        sb.append(indent(indent)).append("]").append(System.lineSeparator());
+        sb.append(indent(indent)).append("]").append(LINE_SEPARATOR);
     }
 
     @Override
@@ -139,7 +147,7 @@ public class JsonDependencyNodeVisitor extends AbstractSerializingVisitor implem
                 .append(value)
                 .append("\"")
                 .append(",")
-                .append(System.lineSeparator());
+                .append(LINE_SEPARATOR);
     }
     /**
      * Appends a key value pair to the string builder without a comma at the end. This is used for the last children of a node.
@@ -158,7 +166,7 @@ public class JsonDependencyNodeVisitor extends AbstractSerializingVisitor implem
                 .append("\"")
                 .append(value)
                 .append("\"")
-                .append(System.lineSeparator());
+                .append(LINE_SEPARATOR);
     }
 
     /**
--- a/src/main/java/org/apache/maven/plugins/dependency/tree/JsonDependencyNodeVisitor.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/tree/JsonDependencyNodeVisitor.java
@@ -1,0 +1,172 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.plugins.dependency.tree;
+
+import java.io.IOException;
+import java.io.Writer;
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.maven.artifact.Artifact;
+import org.apache.maven.shared.dependency.graph.DependencyNode;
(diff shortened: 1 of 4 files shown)


MartinWitt:
CLA is now signed and accepted.

src/main/java/org/apache/maven/plugins/dependency/tree/TreeMojo.java
```diff
@@ -380,6 +380,8 @@ public DependencyNodeVisitor getSerializingDependencyNodeVisitor(Writer writer)
             return new TGFDependencyNodeVisitor(writer);
         } else if ("dot".equals(outputType)) {
             return new DOTDependencyNodeVisitor(writer);
+        } else if ("json".equals(outputType)) {
+            return new JsonDependencyNodeVisitor(writer);
         } else {
             return new SerializingDependencyNodeVisitor(writer, toGraphTokens(tokens));
```
MartinWitt, line 386: In the other PR someone asked for a switch case instead. Shall I convert this to a switch statement?

MartinWitt:
As I am a first-time contributor, someone has to approve the workflow run before we see CI results.

MartinWitt:
@elharo could you approve the worflow run?

src/main/java/org/apache/maven/plugins/dependency/tree/JsonDependencyNodeVisitor.java
```diff
@@ -0,0 +1,180 @@
...
+    /**
+     * Creates a new instance of {@link JsonDependencyNodeVisitor}. The writer will be used to write the output.
+     * @param writer  the writer to write to
+     */
+    public JsonDependencyNodeVisitor(Writer writer) {
+        super(writer);
+    }
+
+    @Override
+    public boolean visit(DependencyNode node) {
```
elharo, line 46: Is it possible for this to et stuck in an infinite recursion with a maliciously hand-crafted dependency node? Or for that matter, with a non-maliciious but buggy circular dependency tree (which does happen)?

src/main/java/org/apache/maven/plugins/dependency/tree/JsonDependencyNodeVisitor.java
```diff
@@ -0,0 +1,180 @@
...
+import org.apache.maven.artifact.Artifact;
+import org.apache.maven.shared.dependency.graph.DependencyNode;
+import org.apache.maven.shared.dependency.graph.traversal.DependencyNodeVisitor;
+
+/**
+ * A dependency node visitor that serializes visited nodes to a writer using the JSON format.
+ */
+public class JsonDependencyNodeVisitor extends AbstractSerializingVisitor implements DependencyNodeVisitor {
+
+    private static final String LINE_SEPARATOR = "\n";
```
elharo, line 34: inline this, a simple "\n" is clearer

src/main/java/org/apache/maven/plugins/dependency/tree/JsonDependencyNodeVisitor.java
```diff
@@ -0,0 +1,180 @@
...
+        sb.append("}").append(LINE_SEPARATOR);
+        try {
+            writer.write(sb.toString());
+        } catch (IOException e) {
+            throw new RuntimeException("Error while writing json output", e);
+            // TODO: handle exception maybe throw runtime exception or mojo exception?
+        }
+    }
+    /**
+     * Appends the node and it's children to the string builder.
```
elharo, line 72: its

src/main/java/org/apache/maven/plugins/dependency/tree/JsonDependencyNodeVisitor.java
```diff
@@ -0,0 +1,180 @@
...
+        }
+    }
+    /**
+     * Appends a key value pair to the string builder.
+     * @param sb  the string builder to append to
+     * @param indent  the current indent level
+     * @param key  the key used as json key
+     * @param value  the value used as json value
+     */
+    private void appendKey(StringBuilder sb, int indent, String key, String value) {
```
elharo, line 139: appendKeyValue

src/main/java/org/apache/maven/plugins/dependency/tree/TreeMojo.java
elharo: no, this is fine

src/test/java/org/apache/maven/plugins/dependency/tree/TestTreeMojo.java
```diff
@@ -127,6 +127,24 @@ public void _testTreeTGFSerializing() throws Exception {
         assertTrue(findString(contents, "testGroupId:release:jar:1.0:compile"));
     }
 
+    /**
+     * Test the JSON format serialization
+     */
+    public void _testTreeJsonSerialzing() throws Exception {
+        List<String> contents = runTreeMojo("tree1.json", "json");
+        assertTrue(findString(contents, "\"testGroupId\": \"project\""));
```
elharo, line 135: not sure why findString is used here. contains is simple and more obvious

MartinWitt: I tried to keep the test case in the same style as the rest of the class. But I can change it.

pombredanne:
@MartinWitt what's left to do to get this through?
FWIW, there is a cottage industry of smalls tools and scripts that are parsing this plugin output .... it would be awesome to have a structure JSON output! 
Anything to help you need move this forward?

src/main/java/org/apache/maven/plugins/dependency/tree/JsonDependencyNodeVisitor.java
```diff
@@ -0,0 +1,179 @@
...
+
+import org.apache.commons.lang3.StringUtils;
+import org.apache.maven.artifact.Artifact;
+import org.apache.maven.shared.dependency.graph.DependencyNode;
+import org.apache.maven.shared.dependency.graph.traversal.DependencyNodeVisitor;
+
+/**
+ * A dependency node visitor that serializes visited nodes to a writer using the JSON format.
+ */
+public class JsonDependencyNodeVisitor extends AbstractSerializingVisitor implements DependencyNodeVisitor {
```
elharo, line 32: Does this need to be public? It's easier to evolve and iterate on if it's not.

src/main/java/org/apache/maven/plugins/dependency/tree/JsonDependencyNodeVisitor.java
```diff
@@ -0,0 +1,179 @@
...
+    private void writeRootNode(DependencyNode node, Writer writer) {
+        int indent = 2;
+        StringBuilder sb = new StringBuilder();
+        sb.append("{").append("\n");
+        writeNode(indent, node, sb);
+        sb.append("}").append("\n");
+        try {
+            writer.write(sb.toString());
+        } catch (IOException e) {
+            throw new RuntimeException("Error while writing json output", e);
```
elharo, line 66: needs a more specific exception. Looking at this now I notice that the superclass and interface are badly designed. They use a PrintWriter which swallows exceptions. That might need to be fixed.

pombredanne: What about just not catch an exception here and leaving this business to the callers.... fixing the PrintWriter could then be done in another patch and not here.

pombredanne:
Looking at the details of this PR, it feels to me that crafting the JSON by hand as done here feels like problems in the making with encoding or else. Should not this use a proper JSON library of sorts? Is not there one bundled in the standard Maven or Java and this plugin may therefore always have have one on hand? 

(Sorry for these likely dumb questions: my Java skills need to brushing off)

yahavi:
If it's useful to anyone, we've developed a Maven dependency tree plugin that generates a JSON dependency tree. Check it out at: https://github.com/jfrog/maven-dep-tree.

monperrus:
@pombredanne @elharo if we are able to drive to merge, we can put more effort on this PR.

src/main/java/org/apache/maven/plugins/dependency/tree/JsonDependencyNodeVisitor.java
elharo: Fixing PrintWriter should be a different PR but that might be a prerequisite for this PR. I'm not sure. It seems fine to let this method throw IOException and handle it further up the stack. RuntimeException is not OK.

src/test/java/org/apache/maven/plugins/dependency/tree/TestTreeMojo.java
elharo: contains is clearer

src/main/java/org/apache/maven/plugins/dependency/tree/JsonDependencyNodeVisitor.java
elharo: add a test for this case, see what works

elharo (changes requested):

src/test/java/org/apache/maven/plugins/dependency/tree/TestTreeMojo.java
LogFlames: I want to double check: do you mean `contents.contains("...")`? 
Due to indentation the string might not be in the list of strings, but a substring of an element, which the `findString`-method seems to be created for in the same file.

LogFlames:
Hi, 
I'm working on getting this merged and incorporating the requested changes. Unfortunately I don't have access to Martin's fork and will need to create a new PR. Will build on Martin's branch and try to update all changes beforehand to make the transition as smooth as possible.

src/test/java/org/apache/maven/plugins/dependency/tree/TestTreeMojo.java
elharo: That makes sense.

jira-importer:
Resolve #1310

PR #327, opened 2023-06-19
Tighten language

elharo:
Trivial change with no effect on build so I can see if the CI is broken at head or not


slawekjaranowski (approved):

elharo:
FYI build does not seem broken at head so something weord happens when updating comons-io to 2.13

PR #328, opened 2023-06-21
MDEP-875 prevent possible NPE

rosti-il:

       pulling in other changes.
       commit message.

### DIFF
--- a/src/main/java/org/apache/maven/plugins/dependency/ListRepositoriesMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/ListRepositoriesMojo.java
@@ -98,14 +98,21 @@ public class ListRepositoriesMojo extends AbstractDependencyMojo {
                 }
             }));
 
+            if (repositories.isEmpty()) {
+                getLog().info("No remote repository is used by this build." + System.lineSeparator());
+                return;
+            }
+
             StringBuilder message = new StringBuilder();
 
             Map<Boolean, List<RemoteRepository>> repoGroupByMirrors = repositories.stream()
                     .collect(Collectors.groupingBy(
                             repo -> repo.getMirroredRepositories().isEmpty()));
 
-            prepareRemoteRepositoriesList(message, repoGroupByMirrors.get(Boolean.TRUE));
-            prepareRemoteMirrorRepositoriesList(message, repoGroupByMirrors.get(Boolean.FALSE));
+            prepareRemoteRepositoriesList(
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/maven/plugins/dependency/ListRepositoriesMojo.java
```diff
@@ -104,8 +104,10 @@ public boolean visitLeave(DependencyNode node) {
                     .collect(Collectors.groupingBy(
                             repo -> repo.getMirroredRepositories().isEmpty()));
 
-            prepareRemoteRepositoriesList(message, repoGroupByMirrors.get(Boolean.TRUE));
-            prepareRemoteMirrorRepositoriesList(message, repoGroupByMirrors.get(Boolean.FALSE));
+            prepareRemoteRepositoriesList(
+                    message, repoGroupByMirrors.getOrDefault(Boolean.TRUE, Collections.emptyList()));
+            prepareRemoteMirrorRepositoriesList(
+                    message, repoGroupByMirrors.getOrDefault(Boolean.FALSE, Collections.emptyList()));
```
slawekjaranowski, line 110: We can also check in methods `prepareRemote*` if list is empty to avoid produce incomplete message.

rosti-il: In case of an empty list is passed to the `prepareRemoteMirrorRepositoriesList()` it doesn't add anything into the `message`, so no check is necessary there. The `prepareRemoteRepositoriesList()` that is called before `prepareRemoteMirrorRepositoriesList()` always appends this title message:
```java
        message.append("Project remote repositories used by this build:").append(System.lineSeparator());
```
Not appending this title message in `prepareRemoteRepositoriesList()` when `prepareRemoteMirrorRepositoriesList()` prints something will look bad, IMHO.

I think a better solution would be checking `repositories.isEmpty()` before line 101 and if it is empty just print some other title message and exit from `doExecute()`. Would you like me to change this pull request this way? What title message for the case of empty `repositories` would you like to be printed? For example: "No remote repository is used by this build."

slawekjaranowski: check in line 101 - ok

> "No remote repository is used by this build." 

will be ok

rosti-il: Done

slawekjaranowski (approved):

PR #374, opened 2024-04-16
[MDEP-917] dependency:analyze-exclusions - use Resolver API instead of ProjectBuilder

slawekjaranowski:
https://issues.apache.org/jira/browse/MDEP-917


jira-importer:
Resolve #1423

PR #378, opened 2024-04-18
[MDEP-922] dependency:analyze-exclusions - should report issue only in current project

slawekjaranowski:
https://issues.apache.org/jira/browse/MDEP-922


jira-importer:
Resolve #1418

PR #380, opened 2024-04-21
[MDEP-914] Fix link in collect goal description

slawekjaranowski:
https://issues.apache.org/jira/browse/MDEP-914


jira-importer:
Resolve #1431

PR #382, opened 2024-04-23
[MDEP-924] Get rid of maven-artifact-transfer from list-classes goal

slawekjaranowski:
https://issues.apache.org/jira/browse/MDEP-924

### DIFF
--- a/src/main/java/org/apache/maven/plugins/dependency/utils/ResolverUtil.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/utils/ResolverUtil.java
@@ -23,14 +23,31 @@ import javax.inject.Named;
 import javax.inject.Provider;
 import javax.inject.Singleton;
 
+import java.util.ArrayList;
 import java.util.Collection;
+import java.util.List;
+import java.util.Objects;
+import java.util.stream.Collectors;
 
 import org.apache.maven.execution.MavenSession;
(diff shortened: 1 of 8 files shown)


cstamas (approved):

src/main/java/org/apache/maven/plugins/dependency/utils/ResolverUtil.java
```diff
@@ -68,4 +85,152 @@ public Collection<Dependency> collectDependencies(Dependency root) throws Depend
...
+        repositoriesList =
+                repositorySystem.newResolutionRepositories(mavenSession.getRepositorySession(), repositoriesList);
+
+        List<RemoteRepository> result = new ArrayList<>(projectRepositories);
+        result.addAll(repositoriesList);
+        return result;
+    }
+
+    // protected for testing purpose
+    protected RemoteRepository prepareRemoteRepository(String repository) {
```
cstamas, line 151: if you resolve/deploy with this repo, you must pass it thru RepoSyste newDeploymentRepository or alike method (to have it mirrors/proxies/auth set).

slawekjaranowski: it is only used for resolving / collecting artifacts

PR #389, opened 2024-05-11
[MDEP-923] Extract copyFile method from AbstractDependencyMojo

slawekjaranowski:
this method is only needed in CopyMojo and CopyDependenciesMojo so it is not needed in AbstractDependencyMojo

### DIFF
--- a/src/main/java/org/apache/maven/plugins/dependency/utils/CopyUtil.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/utils/CopyUtil.java
@@ -1,0 +1,70 @@
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.plugins.dependency.utils;
+
+import javax.inject.Inject;
+import javax.inject.Named;
+import javax.inject.Singleton;
+
+import java.io.File;
+import java.io.IOException;
+
+import org.apache.maven.plugin.MojoExecutionException;
+import org.apache.maven.plugin.logging.Log;
+import org.codehaus.plexus.util.FileUtils;
+import org.sonatype.plexus.build.incremental.BuildContext;
+
+/**
+ * Provide a copyFile method in one place.
+ */
+@Named
+@Singleton
+public class CopyUtil {
+
+    private final BuildContext buildContext;
+
+    @Inject
+    public CopyUtil(BuildContext buildContext) {
+        this.buildContext = buildContext;
+    }
+
+    /**
+     * Does the actual copy of the file and logging.
+     *
+     * @param source represents the file to copy.
+     * @param destination file name of destination file.
+     * @throws MojoExecutionException with a message if an error occurs.
+     */
+    public void copyFile(Log log, File source, File destination) throws MojoExecutionException {
+        try {
+            log.info("Copying " + source + " to " + destination);
+
+            if (source.isDirectory()) {
+                // usual case is a future jar packaging, but there are special cases: classifier and other packaging
+                throw new MojoExecutionException("Artifact has not been packaged yet. When used on reactor artifact, "
+                        + "copy should be executed after packaging: see MDEP-187.");
+            }
+
+            FileUtils.copyFile(source, destination);
+            buildContext.refresh(destination);
+        } catch (IOException e) {
+            throw new MojoExecutionException("Error copying artifact from " + source + " to " + destination, e);
+        }
+    }
+}
--- a/src/main/java/org/apache/maven/plugins/dependency/utils/CopyUtil.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/utils/CopyUtil.java
@@ -33,6 +33,8 @@ import org.sonatype.plexus.build.incremental.BuildContext;
 
 /**
  * Provide a copyFile method in one place.
+ *
+ * @since 3.7.0
  */
 @Named
 @Singleton
@@ -53,6 +55,8 @@ public class CopyUtil {
      * @param source represents the file to copy.
      * @param destination file name of destination file.
      * @throws IOException with a message if an error occurs.
+     *
+     * @since 3.7.0
      */
     public void copyFile(File source, File destination) throws IOException, MojoExecutionException {
         logger.info("Copying {} to {}", source, destination);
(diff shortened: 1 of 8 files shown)


src/main/java/org/apache/maven/plugins/dependency/utils/CopyUtil.java
```diff
@@ -0,0 +1,70 @@
...
+
+    /**
+     * Does the actual copy of the file and logging.
+     *
+     * @param source represents the file to copy.
+     * @param destination file name of destination file.
+     * @throws MojoExecutionException with a message if an error occurs.
+     */
+    public void copyFile(Log log, File source, File destination) throws MojoExecutionException {
+        try {
```
elharo, line 55: I'd strongly prefer to remove the logging completely. 

If it must be logged, it should be debug level only and the logger should be loaded for this class, not passed in.

src/main/java/org/apache/maven/plugins/dependency/utils/CopyUtil.java
```diff
@@ -0,0 +1,70 @@
...
+
+            if (source.isDirectory()) {
+                // usual case is a future jar packaging, but there are special cases: classifier and other packaging
+                throw new MojoExecutionException("Artifact has not been packaged yet. When used on reactor artifact, "
+                        + "copy should be executed after packaging: see MDEP-187.");
+            }
+
+            FileUtils.copyFile(source, destination);
+            buildContext.refresh(destination);
+        } catch (IOException e) {
```
elharo, line 66: Since this isn't a Mojo itself, consider simply letting the IOException bubble up and handle it at a higher level.

src/main/java/org/apache/maven/plugins/dependency/utils/CopyUtil.java
```diff
@@ -0,0 +1,70 @@
...
+import org.apache.maven.plugin.logging.Log;
+import org.codehaus.plexus.util.FileUtils;
+import org.sonatype.plexus.build.incremental.BuildContext;
+
+/**
+ * Provide a copyFile method in one place.
+ */
+@Named
+@Singleton
+public class CopyUtil {
```
elharo, line 38: Since it's on=ly used in this package, nothing here needs to be public.

slawekjaranowski: It is used in another packages of project .... so must be a public 
I can add an interface - which still be a public and implementation which can be package visible, but I'm not sure if needed

src/main/java/org/apache/maven/plugins/dependency/utils/CopyUtil.java
slawekjaranowski: As it is only refactor I don't want to change behavior.

src/main/java/org/apache/maven/plugins/dependency/utils/CopyUtil.java
slawekjaranowski: We will have a `MojoExecutionException` and `IOException` from this method

slawekjaranowski: See line 60 - we have `MojoExecutionException`  - I change it for `IOException` so ww will have one exception from method - @elharo what do you think?

elharo: I'm less sure about that one. That feels close to a MojoExecutionExcepitojn to me. It's OK for on method to throw different exception types for different reasons.

slawekjaranowski: done - please look

src/test/java/org/apache/maven/plugins/dependency/fromDependencies/TestCopyDependenciesMojo.java
```diff
@@ -80,7 +80,7 @@ public void assertNoMarkerFile(Artifact artifact) throws MojoExecutionException
         assertFalse(handle.isMarkerSet());
     }
 
-    public void testCopyFile() throws MojoExecutionException, IOException {
+    public void testCopyFile() throws Exception {
```
elharo, line 83: did this change to throw a raw exception? Otherwise it's better to declare the actual exceptions instead of the common superclass.

src/main/java/org/apache/maven/plugins/dependency/utils/CopyUtil.java
```diff
@@ -0,0 +1,66 @@
...
+
+    /**
+     * Does the actual copy of the file and logging.
+     *
+     * @param source represents the file to copy.
+     * @param destination file name of destination file.
+     * @throws IOException with a message if an error occurs.
+     */
+    public void copyFile(Log log, File source, File destination) throws IOException, MojoExecutionException {
+        log.info("Copying " + source + " to " + destination);
```
elharo, line 55: Log message is likely noise. Either thew copy succeeds and no one wants to read this, ior it fails and an exception is thrown. Demote to debug or remove it.

src/main/java/org/apache/maven/plugins/dependency/utils/CopyUtil.java
```diff
@@ -0,0 +1,66 @@
...
+
+        if (source.isDirectory()) {
+            // usual case is a future jar packaging, but there are special cases: classifier and other packaging
+            throw new MojoExecutionException("Artifact has not been packaged yet. When used on reactor artifact, "
+                    + "copy should be executed after packaging: see MDEP-187.");
```
elharo, line 60: I wonder if we need a URL here? Non-Maven team members might not recognize the issue tracker ID

slawekjaranowski: try paste it in your favorite web search .... 
here similar - it is refactor not improvements

src/main/java/org/apache/maven/plugins/dependency/utils/CopyUtil.java
elharo: YAGNI. I spend way too much time wading through maven logs trying to understand why and how builds failed. Info messages for normal operations just get in the way and make developers' jobs harder. That someone somewhere might one day want this message is not a good enough reason to include it.

I can sort of see letting this slip through as a separate issue from the refactoring, though now that I look at it, I think that refactoring should also remove the log argument here too. This is a new public API we did not have before, and including that Log argument now makes it harder to remove later since it would be a public API change. 

If the choice is between adding this method with the Log argument and not doing the refactoring, I think I'm going to pick not doing the refactoring.

src/test/java/org/apache/maven/plugins/dependency/fromDependencies/TestCopyDependenciesMojo.java
slawekjaranowski: It is a Unit Test ... all other methods simply throws `Exception`

src/main/java/org/apache/maven/plugins/dependency/utils/CopyUtil.java
slawekjaranowski: Log removed from method argument

src/test/java/org/apache/maven/plugins/dependency/fromDependencies/TestCopyDependenciesMojo.java
elharo: I like to know when an exception is added or removed from an API, and I'd make the same comment about the other methods in this class. :-)

elharo (approved):

src/test/java/org/apache/maven/plugins/dependency/fromDependencies/TestCopyDependenciesMojo.java
slawekjaranowski: ok, one exception is new in 3.7.0 but one is old from copied  class - so should I discovered versions for original class or add since now

jira-importer:
Resolve #1420

PR #390, opened 2024-05-11
[MDEP-919] Set m.r.dependencyManagerTransitivity for tree ITs - Maven 4 compatibility

slawekjaranowski:


cstamas (approved):
We should maybe tinker to disable this (by default) instead

jira-importer:
Resolve #1434

PR #391, opened 2024-05-16
[MDEP-799] tree: add optional output type json

LogFlames:
Following this checklist to help us incorporate your 
contribution quickly and easily:

 - [x] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MDEP) filed 
       for the change (usually before you start working on it).  Trivial changes like typos do not 
       require a JIRA issue.  Your pull request should address just this issue, without 
       pulling in other changes.
 - [x] Each commit in the pull request should have a meaningful subject line and body.
 
Same as #325, will squash and update commit in the end, once everything is fixed.
 
 - [x] Format the pull request title like `[MDEP-XXX] - Fixes bug in ApproximateQuantiles`,
       where you replace `MDEP-XXX` with the appropriate JIRA issue. Best practice
       is to use the JIRA issue title in the pull request title and in the first line of the 
       commit message.
 - [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
 - [x] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will 
       be performed on your pull request automatically.
 - [x] You have run the integration tests successfully (`mvn -Prun-its clean verify`).

 - [x] I hereby declare this contribution to be licensed under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

 - [x] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

This is a continuation of #325, which adds JSON output type to the maven-depdenceny-plugin with the goal tree.

The tests were disabled in the previous fork, to enable the tests (in a working state) this is rebased to the current master (2024-05-16), if the history looks weird (have some 'double' commits, unsure how to solve it). Two tests exists.

[Discussion: IOException](https://github.com/apache/maven-dependency-plugin/pull/325#discussion_r1590864832) The JsonDependencyNodeVisitor now uses the PrintWriter which swallows IOExceptions, as the other Node Visitors do. Could not get it to throw an IOException due to the function being defined in the interface without throwing an Exception, unsure if this can be solved without a larger rework?

[...]

PR #398, opened 2024-05-30
[MDEP-799] Add simple info json format for tree and cleanups

slawekjaranowski:
Only simple info for #391 - more detail documentation is still needed

### DIFF
--- a/src/main/java/org/apache/maven/plugins/dependency/tree/TreeMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/tree/TreeMojo.java
@@ -23,7 +23,6 @@ import java.io.IOException;
 import java.io.StringWriter;
 import java.io.Writer;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Objects;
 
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/maven/plugins/dependency/tree/TreeMojo.java
```diff
@@ -133,7 +107,8 @@ public class TreeMojo extends AbstractMojo {
 
     /**
      * If specified, this parameter will cause the dependency tree to be written using the specified format. Currently
-     * supported format are: <code>text</code> (default), <code>dot</code>, <code>graphml</code> and <code>tgf</code>.
+     * supported format are: <code>text</code> (default), <code>dot</code>, <code>graphml</code>, <code>tgf</code>
```
elharo, line 110: format --> formats

src/main/java/org/apache/maven/plugins/dependency/tree/TreeMojo.java
```diff
@@ -86,30 +84,6 @@ public class TreeMojo extends AbstractMojo {
     @Parameter(property = "outputEncoding", defaultValue = "${project.reporting.outputEncoding}")
     private String outputEncoding;
 
-    /**
-     * Contains the full list of projects in the reactor.
-     */
-    @Parameter(defaultValue = "${reactorProjects}", readonly = true, required = true)
```
elharo, line 92: does this cause problems for existing projects that set these?

slawekjaranowski: no problem ... like unused variables are

elharo (approved):

jira-importer:
Resolve #1310

PR #285, opened 2023-01-27
[fix] Correct invalid property name - should be 'overWriteIfNewer'

hazendaz:
was 'overIfNewer'...

note: Have not added Jira yet, looking at a separate issue and this caught my eye as a defect.  Other class in copy dependencies has it spelled correctly but without the prefix mdep.  Tests all pass so nothing was likely using that, I didn't check further at the moment.

Following this checklist to help us incorporate your 
contribution quickly and easily:

 - [ ] Make sure there is a [JIRA issue](https://issues.apache.org/jira/browse/MDEP) filed 
       for the change (usually before you start working on it).  Trivial changes like typos do not 
       require a JIRA issue.  Your pull request should address just this issue, without 
       pulling in other changes.
 - [X] Each commit in the pull request should have a meaningful subject line and body.
 - [ ] Format the pull request title like `[MDEP-XXX] - Fixes bug in ApproximateQuantiles`,
       where you replace `MDEP-XXX` with the appropriate JIRA issue. Best practice
       is to use the JIRA issue title in the pull request title and in the first line of the 
       commit message.
 - [X] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
 - [X] Run `mvn clean verify` to make sure basic checks pass. A more thorough check will 
       be performed on your pull request automatically.
 - [X ] You have run the integration tests successfully (`mvn -Prun-its clean verify`).

If your pull request is about ~20 lines of code you don't need to sign an
[Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf) if you are unsure
please ask on the developers list.

To make clear that you license your contribution under 
the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)
you have to acknowledge this by using the following check-box.

 - [X ] I hereby declare this contribution to be licensed under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0)

 - [ ] In any other case, please file an [Apache Individual Contributor License Agreement](https://www.apache.org/licenses/icla.pdf).

PR #404, opened 2024-06-03
MDEP-938] Correct invalid property name and add deprecated old one

hazendaz:
property = "mdep.overIfNewer" is incorrect as a mismatch on the naming and should have been "mdep.overWriteIfNewer" to respect naming convention as all others do.  All examples show using the more general property as 'overWriteIfNewer' so its odd to pass -Dmdep.overIfNewer instead of expected correct value.  I suspect this is a rare situation and was not an issue that I had but something I noticed when reviewing the code.  The original PR I opened #285 went unmerged and was requested to add a deprecated property for the original along with jira.  Due to how its used, I didn't see a very good way to make the invalid usage populate the correct one.  Maybe there is a better way here.  So for now, the misnamed original 'mdep.overIfNewer' is over a property that is unused simply noting its deprecation.  I added since 3.7.0 which is also odd considering this is simply a fix and the original is entirely still used throughout as it was entirely correct.  Just the property name attribute was incorrect.

I otherwise didn't add any new tests or anything to this situation but could if needed.  Might be better if some error is thrown if the original is even populated as I suspect this along is not enough to indicate it got deprecated since that won't show up in the use case.

       pulling in other changes.
       commit message.

### DIFF
--- a/src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/AbstractFromConfigurationMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/AbstractFromConfigurationMojo.java
@@ -79,8 +79,18 @@ public abstract class AbstractFromConfigurationMojo extends AbstractDependencyMo
      * Overwrite if newer
      *
      * @since 2.0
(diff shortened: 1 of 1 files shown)


slawekjaranowski (approved):

jira-importer:
Resolve #1421

PR #433, opened 2024-10-09
[MDEP-957] By default, don't report slf4j-simple as unused

elharo:
slf4j-simple is a super common "unused" dependency because when slfj-api or slf4j-impl is included and slf4j-simple isn't, the application works just fine but prints an annoying log message to the console telling developers to add it.

```
SLF4J: No SLF4J providers were found.
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See https://www.slf4j.org/codes.html#noProviders for further details.
```

or 
```
SLF4J: Failed to load class "org.slf4j.impl.StaticLoggerBinder".
SLF4J: Defaulting to no-operation (NOP) logger implementation
SLF4J: See https://www.slf4j.org/codes.html#StaticLoggerBinder for further details.
```

so devs deliberately add slf4j-simple to their dependencies just to shut up these warnings.

see https://www.slf4j.org/manual.html

### DIFF
--- a/src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
@@ -228,11 +228,15 @@ public abstract class AbstractAnalyzeMojo extends AbstractMojo {
      * For example, <code>org.apache.*</code> will match all artifacts whose group id starts with
      * <code>org.apache.</code>, and <code>:::*-SNAPSHOT</code> will match all snapshot artifacts.
      * </p>
+     * <p>
+     * By default, org.slf4j:slf4j-simple is ignored. Setting this property to an empty list
+     * will allow it to be detected.
+     * </p>
      *
      * @since 2.10
      */
     @Parameter
-    private String[] ignoredUnusedDeclaredDependencies = new String[0];
+    private String[] ignoredUnusedDeclaredDependencies = {"org.slf4j:slf4j-simple::"};
 
     /**
      * List of dependencies that will be ignored if they are in not test scope but are only used in test classes.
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
```diff
@@ -228,11 +228,15 @@ public abstract class AbstractAnalyzeMojo extends AbstractMojo {
...
+     * <p>
+     * By default, org.slf4j:slf4j-simple is ignored. Setting this property to an empty list
+     * will allow it to be detected.
+     * </p>
      *
      * @since 2.10
      */
     @Parameter
-    private String[] ignoredUnusedDeclaredDependencies = new String[0];
+    private String[] ignoredUnusedDeclaredDependencies = {"org.slf4j:slf4j-simple::"};
```
slawekjaranowski, line 239: 1. why not use a `default` option of parameter?
2. when user defined another ignores list it will be overridden ... maybe we need add it latter to be always ignored ...

elharo: A little later in the file I see this comment

```
    // defaultValue value on @Parameter - not work with Maven 3.2.5
    // When is set defaultValue always win, and there is no possibility to override by plugin configuration.
```

src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
```diff
@@ -228,11 +228,15 @@ public abstract class AbstractAnalyzeMojo extends AbstractMojo {
      * For example, <code>org.apache.*</code> will match all artifacts whose group id starts with
      * <code>org.apache.</code>, and <code>:::*-SNAPSHOT</code> will match all snapshot artifacts.
      * </p>
+     * <p>
+     * By default, org.slf4j:slf4j-simple is ignored. Setting this property to an empty list
+     * will allow it to be detected.
```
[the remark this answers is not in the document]
elharo: I took this comment out. The doc generator will pick up the default value here and document it.

michael-o (approved):
No objection. @slawekjaranowski?

jira-importer:
Resolve #1462

PR #444, opened 2024-10-16
Protect the master branch to avoid accidental commits

elharo:


michael-o (changes requested):
This is something we need to discuss in general AND will this block pushed to master directly? Then no releases will be possible anymore.

elharo:
The goal is to block direct pushes to master. If the release process requires doing that, then yes, we might need to adjust the release process first, which we should do. Direct commits to master are a risk for supply chain attacks. Discussion is ongoing on the dev list.

slachiewicz (approved):

slawekjaranowski (changes requested):
First we need to change a release procedure, scripts before do it.
Now we need direct push to master during release.

slawekjaranowski:
We also should change a rule:
```
The Apache Maven project uses a Commit then Review policy ....
```
https://maven.apache.org/project-roles.html#committers

elharo:
#569

pzygielo:
- apache/maven-site#569

slachiewicz (commented):
-1

PR #445, opened 2024-10-16
Convert to JUnit 5

elharo:
Gets rid of the dependency on commons-io

### DIFF
--- a/src/test/java/org/apache/maven/plugins/dependency/utils/filters/TestDestFileFilter.java
+++ b/src/test/java/org/apache/maven/plugins/dependency/utils/filters/TestDestFileFilter.java
@@ -18,11 +18,6 @@
  */
 package org.apache.maven.plugins.dependency.utils.filters;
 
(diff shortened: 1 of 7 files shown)


src/test/java/org/apache/maven/plugins/dependency/utils/filters/TestDestFileFilter.java
```diff
@@ -23,43 +23,37 @@
...
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertSame;
+import static org.junit.jupiter.api.Assertions.assertTrue;
 
 /**
  * @author brianf
  */
-public class TestDestFileFilter extends TestCase {
+public class TestDestFileFilter {
```
slawekjaranowski, line 43: we can remove `public` classifier for JUnit 5 for class and tests methods

slawekjaranowski (approved):

PR #454, opened 2024-11-09
Use linefeed directly to avoid platform dependent code

elharo:
This plugin seems to have a lot of usage of System.lineSeparator that produces platform dependent output and logs. These days I think pretty much every terminal on every platform should work with a simple \n and when the log goes anywhere other than a terminal then \n works better.

### DIFF
--- a/src/main/java/org/apache/maven/plugins/dependency/ListRepositoriesMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/ListRepositoriesMojo.java
@@ -99,7 +99,7 @@ public class ListRepositoriesMojo extends AbstractDependencyMojo {
             }));
 
             if (repositories.isEmpty()) {
-                getLog().info("No remote repository is used by this build." + System.lineSeparator());
+                getLog().info("No remote repository is used by this build.\n");
                 return;
             }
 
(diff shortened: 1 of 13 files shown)


michael-o (changes requested):
Don't make assumptions about a target.

slawekjaranowski (changes requested):
I don't see a benefit of it ... but can introduce a problems

elharo:
The benefit is that the code does exactly the same thing and produces the same output on different platforms. This is especially important when writing files since files are not necessarily read on the same platform where they're written.

It's also simpler to write \n instead of System.lineSeparator(). Lines are shorter and it's less code to read.

elharo:
For example, consider this fragment from JavadocUtiltest:

```
        // Other tests
        version = "java full version \"1.5.0_07-164\"" + System.getProperty("line.separator");
        assertEquals("1.5.0", JavadocUtil.extractJavadocVersion(version));

        version = System.getProperty("line.separator") + "java full version \"1.5.0_07-164\"";
        assertEquals("1.5.0", JavadocUtil.extractJavadocVersion(version));

        version = System.getProperty("line.separator") + "java full version \"1.5.0_07-164\""
                + System.getProperty("line.separator");
        assertEquals("1.5.0", JavadocUtil.extractJavadocVersion(version));

        version = "java full" + System.getProperty("line.separator") + " version \"1.5.0_07-164\"";
        assertEquals("1.5.0", JavadocUtil.extractJavadocVersion(version));
```

Here we test different strings depending on which platform the test runs on. We should instead run the same tests on all platforms.

PR #457, opened 2024-11-10
Stop setting deprecated dependency artifacts

elharo:


Bukama (approved):
Deleted code is the best code :D

Non-binding approval by me.

slawekjaranowski (approved):

PR #458, opened 2024-11-10
Remove unsupported goals

elharo:
maven-invoker-plugin doesn't have these goals. probably copy pasta

https://maven.apache.org/plugins/maven-invoker-plugin/plugin-info.html


pzygielo:
These are not supposed to be goals of m-invoker-p, but `goals` is rather [configuration element](https://maven.apache.org/plugins/maven-invoker-plugin/run-mojo.html#goals) I guess.

PR #459, opened 2024-11-11
Allowlist org.glassfish:javax.json

elharo:


src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
```diff
@@ -231,7 +231,7 @@ public abstract class AbstractAnalyzeMojo extends AbstractMojo {
      *
      * @since 2.10
      */
-    @Parameter(defaultValue = "org.slf4j:slf4j-simple::")
+    @Parameter(defaultValue = "org.slf4j:slf4j-simple::,org.glassfish:javax.json::")
```
slachiewicz, line 234: How big could this list of `hidden` default exclude be? and why this javax.json specially?

elharo: Probably less than 100 over time. Each one is an improvement.

javax.json because it's the next one I happened to stumble over when working on our own code.

slawekjaranowski: We should put it in other list ... when we use it as default it will be overridden when user use own list.

Should be added always, maybe we need a new parameter like `useDefaultExcludes` or some else ...

slawekjaranowski: I'm not sure if we need add every dependencies which we need in a specific projects ....
In such way we will have very long list

slawekjaranowski (changes requested):
We should put it in other list ... when we use it as default it will be overridden when user use own list.

src/main/java/org/apache/maven/plugins/dependency/analyze/AbstractAnalyzeMojo.java
elharo: I don't think we will have a very long list. ChatGPT-4 has roughly 1.8 *trillion* parameters. In 2025 that's a long list. Here, there might be a few dozen special cases occasionally encountered, though unless someone spends a lot of time on this, we will probably include less than ten. This is a *small* list.

slawekjaranowski: nevermind how list is long will be lost when user add own excludes, so we should implement it in other way

elharo: What do you suggest? We could include a base list of dependencies we always ignore and allow the user to add to this list.

slawekjaranowski: > What do you suggest? We could include a base list of dependencies we always ignore and allow the user to add to this list.

exactly - user should have a possibility to not use default excluded list

elharo: Do you want the user to replace the default list then, not simply append to it?

My gut is that we should not warn on anything we're not sure about, and we're never sure about dependencies like slf4j that are commonly used by reflection.

elharo: OK, so let me take a run at Option 1 by adding code that enables clients to add to this list. Meanwhile, this PR can be approved since org.glassfish:javax.json is one of those dwpendencies that shouldn't be warned about.

elharo:
org.glassfish:javax.json is not private project specific. This is a widely used dependency that shows up in a lot of trees. It ended up in the glassfish project for no particularly good reason. There are other implementations floating around we'll likely need to allow list too, but this is a very commonly encountered one.

PR #465, opened 2024-11-13
Add additional comment to clarify the minimal supported version of ou…

gluxhappy:
…tputing dependen

### DIFF
--- a/src/main/java/org/apache/maven/plugins/dependency/tree/TreeMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/tree/TreeMojo.java
@@ -108,7 +108,7 @@ public class TreeMojo extends AbstractMojo {
     /**
      * If specified, this parameter will cause the dependency tree to be written using the specified format. Currently
      * supported formats are: <code>text</code> (default), <code>dot</code>, <code>graphml</code>, <code>tgf</code>
-     * and <code>json</code>(since 3.7.0).
+     * and <code>json</code> (since 3.7.0).
      * These additional formats can be plotted to image files.
      *
      * @since 2.2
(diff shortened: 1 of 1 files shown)


src/main/java/org/apache/maven/plugins/dependency/tree/TreeMojo.java
```diff
@@ -108,7 +108,7 @@ public class TreeMojo extends AbstractMojo {
     /**
      * If specified, this parameter will cause the dependency tree to be written using the specified format. Currently
      * supported formats are: <code>text</code> (default), <code>dot</code>, <code>graphml</code>, <code>tgf</code>
-     * and <code>json</code>.
+     * and <code>json</code>(since 3.7.0).
```
elharo, line 111: Needs a space

elharo (commented):
For myself, I prefer to focus the docs on the current version. I really don't care what was true several versions ago.

src/main/java/org/apache/maven/plugins/dependency/tree/TreeMojo.java
gluxhappy: The "Since" is already everywhere. 😂

src/main/java/org/apache/maven/plugins/dependency/tree/TreeMojo.java
```diff
@@ -108,7 +108,7 @@ public class TreeMojo extends AbstractMojo {
     /**
      * If specified, this parameter will cause the dependency tree to be written using the specified format. Currently
      * supported formats are: <code>text</code> (default), <code>dot</code>, <code>graphml</code>, <code>tgf</code>
-     * and <code>json</code>.
+     * and <code>json</code> (since 3.7.0).
```
elharo, line 111: I don't know that we need this. It's implicitly the current version being described.

slawekjaranowski: It is information in which version such feature was added - for me it is ok

slawekjaranowski (approved):

PR #469, opened 2024-11-16
[MDEP-952] Rollback bad commit

elharo:
Bisecting pointed to the commit I'm rolling back here, and I've manually verified that this fixes the bug. After this change goes in, I plan to see if I can construct an automated integration test that exposes the bug. Then, we can start rolling these changes forward more carefully one at a time until we isolate the problem. Since the original change went in without PR or code review, I think this is acceptable.


pzygielo:
f64df8fd3f2f295312c9f10169585705c75bf03d

slawekjaranowski (approved):

PR #471, opened 2024-11-16
[MDEP-966] Add previously undeclared but used sisu dependency

elharo:


Bukama (approved):
I had in mind that there is an enforce rule that ensures to all dependencies must be defined explicitly, but I don't find it (or I remember wrong). But so from my amateurish (in this context) point of view this explicit declaration is the way we should do it.

so non binding approve by me.

pom.xml
```diff
@@ -171,6 +171,11 @@ under the License.
       <artifactId>plexus-archiver</artifactId>
       <version>${plexus-archiver.version}</version>
     </dependency>
+    <dependency>
+      <groupId>org.eclipse.sisu</groupId>
+      <artifactId>org.eclipse.sisu.plexus</artifactId>
+      <version>0.9.0.M3</version>
+    </dependency>
```
slawekjaranowski, line 178: please check if it is provided by Maven

elharo: done

slachiewicz (approved):

jira-importer:
Resolve #1480

PR #478, opened 2024-11-19
[MDEP-952] Break dependency on commons-lang3

elharo:
This relies on getIncludes and getExcludes never returning null, which is currently true. Since this is all in src/test this reasonably safe and should have no effect on production.


michael-o (approved):

jira-importer:
Resolve #1478

PR #482, opened 2024-11-19
Remove duplicate dependency definition

elharo:


gnodet:
That one is also in #480 afaik

gnodet (approved):

PR #483, opened 2024-11-19
[MDEP-966] Migrate TreeMojo/CopyMojo/AnalyzeExclusionsMojo/UnpackMojo/CopyDependenciesMojo from Plexus to Sisu Guice

elharo:


slawekjaranowski (approved):

src/main/java/org/apache/maven/plugins/dependency/exclusion/AnalyzeExclusionsMojo.java
```diff
@@ -57,15 +58,19 @@
 @Mojo(name = "analyze-exclusions", requiresDependencyCollection = ResolutionScope.TEST, threadSafe = true)
 public class AnalyzeExclusionsMojo extends AbstractMojo {
 
-    @Component
     private MavenProject project;
```
gnodet, line 61: Constructor injected fields should be made `final`.

gnodet (changes requested):

src/main/java/org/apache/maven/plugins/dependency/exclusion/AnalyzeExclusionsMojo.java
elharo: good idea, done

jira-importer:
Resolve #1480

PR #485, opened 2024-11-20
[MDEP-952] avoid StringUtils.stripEnd

elharo:
In this case the simplest thing that could possibly work is to simply not strip the colon from the end. It's just a slightly uglier log message.


slachiewicz (approved):

jira-importer:
Resolve #1478

PR #487, opened 2024-11-20
[MDEP-952] Partially decouple AbstractFromConfigurationMojo from StringUtils

elharo:
Needs a closer look. This appears to be the change that caused the bug MDEP-952

```
Running post-build script: /Users/elharo/maven-dependency-plugin/target/it/mdep-952-unpack-fails-if-extension-of-artifact-is-used/verify.groovy
Assertion failed: 

assert buildLog.contains( 'BUILD SUCCESS' )
       |        |
       |        false
```

### DIFF
--- a/src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/AbstractFromConfigurationMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/AbstractFromConfigurationMojo.java
@@ -181,10 +181,6 @@ public abstract class AbstractFromConfigurationMojo extends AbstractDependencyMo
             } catch (ArtifactFilterException e) {
                 throw new MojoExecutionException(e.getMessage(), e);
             }
-
-            if (artifactItem.getType() == null) {
-                throw new NullPointerException("Missing type for " + artifactItem.getArtifactId());
-            }
         }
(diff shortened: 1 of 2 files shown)


src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/AbstractFromConfigurationMojo.java
```diff
@@ -367,7 +367,7 @@ public void setArtifact(String artifact) throws MojoFailureException {
         if (artifact != null) {
             String packaging = "jar";
             String classifier;
-            String[] tokens = StringUtils.split(artifact, ":");
+            String[] tokens = artifact.split(":");
```
elharo, line 370: This is likely the issue. I think the split methods may not be quite the same.

elharo:
Hooray for tests. I was able to narrow down the problem to the change from StringUtils.split(foo to foo.split.

I'm not immediately sure why/how those two split methods differ but they do. So this PR now removes usages of StringUtils except for StringUtils.split. I'll address that carefully in a subsequent PR after more investigation.

slachiewicz (approved):

jira-importer:
Resolve #1478

PR #489, opened 2024-11-23
Avoid deprecated ArtifactFactory

elharo:


slawekjaranowski (approved):

src/test/java/org/apache/maven/plugins/dependency/fromDependencies/TestCopyDependenciesMojo2.java
```diff
@@ -246,11 +243,12 @@ public void testRepositoryLayout() throws Exception {
         File baseDirectory = Paths.get(targetRepository.getBasedir()).toFile();
         assertTrue(baseDirectory.isDirectory());
 
+        org.apache.maven.repository.RepositorySystem repositorySystem =
+                lookup(org.apache.maven.repository.RepositorySystem.class);
```
slawekjaranowski, line 247: grr `org.apache.maven.repository.RepositorySystem` is also deprecated , it has only Legacy implementation ... 😄

elharo: Bleah, still trying to figure out what needs to be done for these sorts of thing. It would be too easy if whoever deprecated this way back when had documented the replacement.

slawekjaranowski: 😄 our deprecated code is not well documented ...
In this case you can simply use `new DefaultArtifact(....)`

PR #521, opened 2025-04-19
Fix broken link for analyze-exclusions-mojo on usage-page

Bukama:
There were actual two links for the same goal, one with the correct link, but at the bottom of the page.
The broken one was added later in "alphabetical correct" position.

I fixed the one at the top where it belongs and removed the one at the bottom with


Bukama:
related issue (in maven site) https://github.com/apache/maven-site/issues/760

slawekjaranowski (approved):

PR #522, opened 2025-04-19
Enable GH issues

Bukama:


slawekjaranowski:
We should update a current PR template:

https://github.com/apache/maven-dependency-plugin/blob/master/.github/pull_request_template.md

to something like:
https://github.com/apache/maven-jar-plugin/blob/master/.github/pull_request_template.md

.github/workflows/release-drafter.yml
```diff
@@ -22,6 +22,7 @@ on:
   push:
     branches:
       - master
+workflow_dispatch:
```
pzygielo, line 25: - https://github.com/apache/maven-javadoc-plugin/pull/365#discussion_r2051455215

Bukama:
> We should update a current PR template:
> 
> https://github.com/apache/maven-dependency-plugin/blob/master/.github/pull_request_template.md
> 
> to something like: https://github.com/apache/maven-jar-plugin/blob/master/.github/pull_request_template.md

replaced with the one from jar-plugin

.github/workflows/release-drafter.yml
Bukama: > * [Enable GH issues maven-javadoc-plugin#365 (comment)](https://github.com/apache/maven-javadoc-plugin/pull/365#discussion_r2051455215)

fixed

slawekjaranowski (approved):

PR #532, opened 2025-05-27
[MDEP-979] Clean up AbstractDependencyMojoTestCase setUp

Ndacyayisenga-droid:

Fixes https://issues.apache.org/jira/browse/MDEP-979

### DIFF
--- a/src/test/java/org/apache/maven/plugins/dependency/AbstractDependencyMojoTestCase.java
+++ b/src/test/java/org/apache/maven/plugins/dependency/AbstractDependencyMojoTestCase.java
@@ -35,26 +35,49 @@ import org.eclipse.aether.RepositorySystem;
 import org.eclipse.aether.RepositorySystemSession;
 import org.eclipse.aether.repository.LocalRepository;
 import org.eclipse.aether.repository.LocalRepositoryManager;
+import org.junit.Before;
 import org.sonatype.plexus.build.incremental.DefaultBuildContext;
 
 public abstract class AbstractDependencyMojoTestCase extends AbstractMojoTestCase {
 
     protected File testDir;
-
     protected DependencyArtifactStubFactory stubFactory;
 
-    protected void setUp(String testDirectoryName, boolean createFiles) throws Exception {
-        setUp(testDirectoryName, createFiles, true);
-    }
-
-    protected void setUp(String testDirectoryName, boolean createFiles, boolean flattenedPath) throws Exception {
-        // required for mojo lookups to work
+    @Before
+    public void setUp() throws Exception {
+        // Initialize mojo lookups as required by AbstractMojoTestCase
         super.setUp();
 
+        // Create temporary test directory
+        testDir = Files.createTempDirectory("test-dependency").toFile();
+        testDir.deleteOnExit();
+
+        // Initialize stub factory with default settings
+        stubFactory = new DependencyArtifactStubFactory(testDir, true, true);
+    }
+
+    /**
+     * Allows subclasses to customize the setup with specific test directory name and stub factory settings.
+     *
+     * @param testDirectoryName The name for the temporary test directory.
+     * @param createFiles Whether to create files in the stub factory.
+     * @param flattenedPath Whether to use flattened paths in the stub factory.
+     * @throws Exception If setup fails.
+     */
+    protected void customizeSetUp(String testDirectoryName, boolean createFiles, boolean flattenedPath)
+            throws Exception {
+        // Clean up existing test directory if present
+        if (testDir != null) {
+            FileUtils.deleteDirectory(testDir);
+            assertFalse(testDir.exists());
+        }
+
+        // Create new test directory with specified name
         testDir = Files.createTempDirectory(testDirectoryName).toFile();
         testDir.deleteOnExit();
 
-        stubFactory = new DependencyArtifactStubFactory(this.testDir, createFiles, flattenedPath);
+        // Reinitialize stub factory with custom settings
+        stubFactory = new DependencyArtifactStubFactory(testDir, createFiles, flattenedPath);
     }
 
     @Override
--- a/src/test/java/org/apache/maven/plugins/dependency/AbstractDependencyMojoTestCase.java
+++ b/src/test/java/org/apache/maven/plugins/dependency/AbstractDependencyMojoTestCase.java
@@ -21,6 +21,7 @@ package org.apache.maven.plugins.dependency;
 import java.io.File;
 import java.io.IOException;
 import java.nio.file.Files;
+import java.util.UUID;
 
 import org.apache.commons.io.FileUtils;
 import org.apache.maven.artifact.Artifact;
@@ -48,8 +49,9 @@ public abstract class AbstractDependencyMojoTestCase extends AbstractMojoTestCas
         // Initialize mojo lookups as required by AbstractMojoTestCase
         super.setUp();
 
-        // Create temporary test directory
-        testDir = Files.createTempDirectory("test-dependency").toFile();
+        // Create a unique temporary test directory to avoid parallel test conflicts
+        String uniqueDirName = "test-dependency" + UUID.randomUUID();
+        testDir = Files.createTempDirectory(uniqueDirName).toFile();
         testDir.deleteOnExit();
 
         // Initialize stub factory with default settings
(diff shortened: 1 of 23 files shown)


src/test/java/org/apache/maven/plugins/dependency/AbstractDependencyMojoTestCase.java
```diff
@@ -35,26 +35,49 @@
...
+     *
+     * @param testDirectoryName The name for the temporary test directory.
+     * @param createFiles Whether to create files in the stub factory.
+     * @param flattenedPath Whether to use flattened paths in the stub factory.
+     * @throws Exception If setup fails.
+     */
+    protected void customizeSetUp(String testDirectoryName, boolean createFiles, boolean flattenedPath)
+            throws Exception {
+        // Clean up existing test directory if present
+        if (testDir != null) {
```
elharo, line 70: This could be very wonky when  tests are run in parallel. Tests should clean up after themselves, not when the next test method is set up.

src/test/java/org/apache/maven/plugins/dependency/AbstractDependencyMojoTestCase.java
```diff
@@ -35,26 +35,49 @@
...
-        super.setUp();
+    /**
+     * Allows subclasses to customize the setup with specific test directory name and stub factory settings.
+     *
+     * @param testDirectoryName The name for the temporary test directory.
+     * @param createFiles Whether to create files in the stub factory.
+     * @param flattenedPath Whether to use flattened paths in the stub factory.
+     * @throws Exception If setup fails.
+     */
+    protected void customizeSetUp(String testDirectoryName, boolean createFiles, boolean flattenedPath)
```
elharo, line 67: Is this meant to be invoked or overridden? If the former, then make it final. But consider calling this from setUp and letting subclasses override it instead.

src/test/java/org/apache/maven/plugins/dependency/AbstractDependencyMojoTestCase.java
```diff
@@ -35,26 +35,49 @@
...
+        stubFactory = new DependencyArtifactStubFactory(testDir, true, true);
     }
 
-    protected void setUp(String testDirectoryName, boolean createFiles, boolean flattenedPath) throws Exception {
-        // required for mojo lookups to work
-        super.setUp();
+    /**
+     * Allows subclasses to customize the setup with specific test directory name and stub factory settings.
+     *
+     * @param testDirectoryName The name for the temporary test directory.
```
elharo, line 62: the name for the temporary test directory

src/test/java/org/apache/maven/plugins/dependency/fromConfiguration/TestArtifactItem.java
```diff
@@ -22,11 +22,15 @@
...
 import org.apache.maven.artifact.Artifact;
 import org.apache.maven.plugins.dependency.AbstractDependencyMojoTestCase;
+import org.junit.Before;
 
 public class TestArtifactItem extends AbstractDependencyMojoTestCase {
 
-    protected void setUp() throws Exception {
-        setUp("artifactItems", false);
+    @Before
+    public void setUp() throws Exception {
```
elharo, line 30: @Override

src/test/java/org/apache/maven/plugins/dependency/fromDependencies/TestCopyDependenciesMojo2.java
```diff
@@ -46,15 +46,19 @@
...
 public class TestCopyDependenciesMojo2 extends AbstractDependencyMojoTestCase {
 
     private CopyDependenciesMojo mojo;
 
-    @Override
-    protected void setUp() throws Exception {
-        // required for mojo lookups to work
-        super.setUp("copy-dependencies", true);
+    @Before
+    public void setUp() throws Exception {
```
elharo, line 56: @Override

src/test/java/org/apache/maven/plugins/dependency/utils/TestDependencyStatusSets.java
```diff
@@ -19,12 +19,15 @@
...
 import org.apache.maven.plugins.dependency.AbstractDependencyMojoTestCase;
+import org.junit.Before;
 
 public class TestDependencyStatusSets extends AbstractDependencyMojoTestCase {
 
-    protected void setUp() throws Exception {
-        // required for mojo lookups to work
-        super.setUp("dss", true);
+    @Before
+    public void setUp() throws Exception {
```
elharo, line 27: @Override

src/test/java/org/apache/maven/plugins/dependency/utils/translators/TestClassifierTypeTranslator.java
```diff
@@ -48,9 +49,11 @@ public class TestClassifierTypeTranslator extends AbstractDependencyMojoTestCase
 
     private ArtifactHandlerManager artifactHandlerManager;
 
-    @Override
-    protected void setUp() throws Exception {
-        super.setUp("classifiertype-translator", false);
+    @Before
+    public void setUp() throws Exception {
```
elharo, line 53: @Override

elharo (changes requested):

src/test/java/org/apache/maven/plugins/dependency/AbstractDependencyMojoTestCase.java
```diff
@@ -35,33 +36,52 @@
...
-    protected void setUp(String testDirectoryName, boolean createFiles, boolean flattenedPath) throws Exception {
-        // required for mojo lookups to work
+    @Override
+    @Before
+    public void setUp() throws Exception {
+        // Initialize mojo lookups as required by AbstractMojoTestCase
         super.setUp();
 
+        // Create a unique temporary test directory to avoid parallel test conflicts
+        String uniqueDirName = "test-dependency" + UUID.randomUUID();
```
elharo, line 54: I don't think you need UUIDs here. JUnit can create temporary directories with @TempDir. Or do we not have JUnit 5 here? Either way, let's makes sure we only create one temporary directory per test.

src/test/java/org/apache/maven/plugins/dependency/AbstractDependencyMojoTestCase.java
```diff
@@ -35,33 +36,52 @@
...
+        testDir.deleteOnExit();
+
+        // Initialize stub factory with default settings
+        stubFactory = new DependencyArtifactStubFactory(testDir, true, true);
+    }
+
+    /**
+     * Allows subclasses to customize the setup with specific test directory name and stub factory settings.
+     *
+     * @param testDirectoryName the name for the temporary test directory.
```
elharo, line 65: no period

src/test/java/org/apache/maven/plugins/dependency/AbstractDependencyMojoTestCase.java
```diff
@@ -35,33 +36,52 @@
...
+     * @param createFiles whether to create files in the stub factory.
+     * @param flattenedPath whether to use flattened paths in the stub factory.
+     * @throws Exception If setup fails.
+     */
+    protected final void customizeSetUp(String testDirectoryName, boolean createFiles, boolean flattenedPath)
```
elharo, line 70: Probably setUp should be calling this method instead of duplicating this code

src/test/java/org/apache/maven/plugins/dependency/TestCollectMojo.java
```diff
@@ -29,13 +29,17 @@
...
 
     @Override
-    protected void setUp() throws Exception {
-        // required for mojo lookups to work
-        super.setUp("markers", false);
+    @Before
+    public void setUp() throws Exception {
+        // Call superclass setup (initializes mojo lookups and default test directory)
+        super.setUp();
+        customizeSetUp("markers", false, true);
```
elharo, line 41: not needed, super.setUp should invoke this method which can be overridden here.

elharo (changes requested):

src/test/java/org/apache/maven/plugins/dependency/AbstractDependencyMojoTestCase.java
```diff
@@ -35,33 +36,52 @@
...
     protected DependencyArtifactStubFactory stubFactory;
 
-    protected void setUp(String testDirectoryName, boolean createFiles) throws Exception {
-        setUp(testDirectoryName, createFiles, true);
-    }
-
-    protected void setUp(String testDirectoryName, boolean createFiles, boolean flattenedPath) throws Exception {
-        // required for mojo lookups to work
+    @Override
+    @Before
```
pzygielo, line 48: This is misleading.
This is still JUnit3 TestCase (and its successors are executed as such) and JUnit's annotations have no power here (or there).

src/test/java/org/apache/maven/plugins/dependency/AbstractDependencyMojoTestCase.java
pzygielo: > JUnit can create temporary directories with `@Tempdir`.

I'm not sure `@Tempdir` would work here, for the same reason as in https://github.com/apache/maven-dependency-plugin/pull/532/files#r2115815689.

> Or do we not have JUnit 5 here?

It's not enough in test-time.

src/test/java/org/apache/maven/plugins/dependency/AbstractDependencyMojoTestCase.java
pzygielo: - https://github.com/apache/maven-javadoc-plugin/pull/346#issuecomment-2521034423

jira-importer:
Resolve #1486

Ndacyayisenga-droid:
@elharo sadly, I tried this again on a different branch on my local. I close this PR and open a new one

Issue #1486, opened 2025-04-18
[MDEP-979] Clean up AbstractDependencyMojoTestCase setUp


jira-importer:
**[Elliotte Rusty Harold](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=elharo)** opened **[MDEP-979](https://issues.apache.org/jira/browse/MDEP-979?redirect=false)** and commented

The setUp methods here are a mess and do not follow JUnit conventions. Furthermore, subclasses do not always properly invoke super.setUp

---

**Remote Links:**
- [GitHub Pull Request #532
](https://github.com/apache/maven-dependency-plugin/pull/532)

PR #1490, opened 2025-06-26
Clean up AbstractDependencyMojoTestCase setUp

Ndacyayisenga-droid:

Fixes https://github.com/apache/maven-dependency-plugin/issues/1486


elharo (changes requested):
I think this one needs the ICKA signed or a commitment to license under apache with the checkbox, or something

olamy:
> I think this one needs the ICKA signed or a commitment to license under apache with the checkbox, or something

Not sure as most of the code is for src/test/. 
But why ICLA if @Ndacyayisenga-droid  is happy to do it. See here https://www.apache.org/licenses/contributor-agreements.html#submitting

Issue #1458, opened 2024-11-09
[MDEP-964] Allowlist org.glassfish:javax.json


jira-importer:
**[Elliotte Rusty Harold](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=elharo)** opened **[MDEP-964](https://issues.apache.org/jira/browse/MDEP-964?redirect=false)** and commented

found this one in our own code:

- 

\<dependency>
- 

      <groupId>org.glassfish</groupId>

- 

      <artifactId>javax.json</artifactId>

- 

      <version>1.1.4</version>

- 

      <scope>test</scope>

- 

\</dependency>
It's typically loaded by reflection so not found by the analyzer. Might want to list it as used iff javax.json-api is used

---

**Remote Links:**
- [GitHub Pull Request #459
](https://github.com/apache/maven-dependency-plugin/pull/459)

PR #1492, opened 2025-07-07
[MEDP-964] unconditionally ignore dependencies known to be loaded by reflection

elharo:
see https://github.com/apache/maven-dependency-plugin/pull/459 and #1458


slawekjaranowski:
The site to update https://maven.apache.org/plugins/maven-dependency-plugin/examples/exclude-dependencies-from-dependency-analysis.html

We should add info about items ignored by default

PR #1517, opened 2025-09-13
Improve usage of silent parameter in Copy and Unpack goals

slawekjaranowski:
- parameter should be defined where is used, so move to child class from root class AbstractDependencyMojo
- use it during logging of coping or unpacking


src/main/java/org/apache/maven/plugins/dependency/utils/CopyUtil.java
```diff
@@ -55,20 +55,27 @@ public CopyUtil(BuildContext buildContext) {
      *
      * @param sourceArtifact the artifact (file) to copy
      * @param destination file name of destination file
+     * @param silent if true, use debug log level, otherwise info for logging
```
elharo, line 58: This feels wrong. Logging should be controlled by log levels, not Maven parameters

src/main/java/org/apache/maven/plugins/dependency/utils/CopyUtil.java
```diff
@@ -78,12 +85,15 @@ public void copyArtifactFile(Artifact sourceArtifact, File destination) throws I
      *
      * @param source the source file to copy
      * @param destination the destination file
+     * @param silent if true, use debug log level, otherwise info for logging
      * @throws IOException if copy has failed
      *
      * @since 3.2.0
      */
-    public void copyFile(File source, File destination) throws IOException {
-        logger.debug("Copying file '{}' to {}", source, destination);
```
elharo, line 86: I really can't see why this should be anything other than debug by default. This is not an actionable message.

src/main/java/org/apache/maven/plugins/dependency/utils/CopyUtil.java
```diff
@@ -78,12 +85,15 @@ public void copyArtifactFile(Artifact sourceArtifact, File destination) throws I
...
+    public void copyFile(File source, File destination, boolean silent) throws IOException {
+        if (!silent) {
+            logger.info("Copying file '{}' to {}", source, destination);
```
elharo, line 95: and now if the user turns on debugging but this is silent they don't see this

src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/AbstractFromConfigurationMojo.java
```diff
@@ -92,6 +92,15 @@ public abstract class AbstractFromConfigurationMojo extends AbstractDependencyMo
     @Parameter(property = "mdep.overIfNewer", defaultValue = "true")
     private boolean overIfNewer;
 
+    /**
+     * If the plugin should be more silent with logging.
+     * <br/>
+     * Use {@code -q} command line option if you want to suppress all output.
+     * @since 2.0
+     */
+    @Parameter(property = "silent", defaultValue = "false")
```
elharo, line 101: This should still be deprecated as it was where it was moved from

elharo (changes requested):

src/main/java/org/apache/maven/plugins/dependency/utils/CopyUtil.java
slawekjaranowski: eg. When using `copy` mojo from cli I need an information what was copied for auditing

elharo:
related: https://github.com/apache/maven-dependency-plugin/pull/447

elharo:
I do not want a large set of switches for every imaginable combination of logging this and that but not some other thing. The default is not to log things users don;t need to know and aren't going to act on the vast majority of the time. If you need more than that, turn on the debug logs. Yes, it's verbose but so is logging every file that's copied.

slawekjaranowski:
> I do not want a large set of switches for every imaginable combination of logging this and that but not some other thing. 

so I moved it where is used ... we have small number of switches

> The default is not to log things users don;t need to know and aren't going to act on the vast majority of the time. 

In this case I think it is useful information what happens

> If you need more than that, turn on the debug logs. Yes, it's verbose but so is logging every file that's copied.

Similar we can say, if you don't want logs use -q options

elharo:
Yes, exactly. Silent is debug, We don't need anything else. Our logs are vastly too noisy with severe consequences for usability.

PR #1566, opened 2025-12-19
Cleaner exception handling

elharo:


slawekjaranowski (approved):

rmannibucau:
hmm, why this broken pattern got merged? we shouldnt encourage to read before closing IO IMHO

elharo:
I stand by this PR. I think this is correct. If you can construct an example that shows a failure here. I'll reconsider.

rmannibucau:
@elharo I gave why I think it is worse than previous code by design there https://github.com/apache/maven-dependency-plugin/pull/1566#issuecomment-3675793769

it is not intended to use the finalized value before close is called and new code does

PR #1571, opened 2025-12-29
Introduce graphRoots for dependencyFilter based mojos

rfscholte:

- [X] Your pull request should address just one issue, without pulling in other changes.
- [X] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
- [X] Run `mvn verify` to make sure basic checks pass.
- [X] You have run the integration tests successfully (`mvn -Prun-its verify`).

### DIFF
--- a/src/main/java/org/apache/maven/plugins/dependency/fromDependencies/AbstractDependencyFilterMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/fromDependencies/AbstractDependencyFilterMojo.java
@@ -529,8 +529,7 @@ public abstract class AbstractDependencyFilterMojo extends AbstractDependencyMoj
         org.eclipse.aether.graph.Dependency dependency = RepositoryUtils.toDependency(
                 root, session.getRepositorySession().getArtifactTypeRegistry());
 
-        List<RemoteRepository> remoteRepositories =
-                RepositoryUtils.toRepos(session.getProjectBuildingRequest().getRemoteRepositories());
+        List<RemoteRepository> remoteRepositories = getProject().getRemoteProjectRepositories();
 
         Collection<org.eclipse.aether.artifact.Artifact> depArtifacts =
                 resolverUtil.resolveDependencies(dependency.getArtifact(), remoteRepositories);


src/main/java/org/apache/maven/plugins/dependency/fromDependencies/AbstractDependencyFilterMojo.java
```diff
@@ -231,6 +237,16 @@ public abstract class AbstractDependencyFilterMojo extends AbstractDependencyMoj
...
 
+    /**
+     * By default this matches the project itself is the root.
+     * With graphRoots you can choose for a subtree of dependencies.
+     * After that the general include/exclude filters can be applied.
+     *
+     * @since 3.10.0
+     */
+    @Parameter
+    private List<GraphRoot> graphRoots;
```
mthmulders: Given that `graphRoots` is a list, wouldn't it make sense to have an IT that verifies the collection of dependencies with *multiple* `graphRoots` defined?

src/main/java/org/apache/maven/plugins/dependency/fromDependencies/AbstractDependencyFilterMojo.java
```diff
@@ -479,6 +502,38 @@ private Set<Artifact> resolve(Set<org.eclipse.aether.artifact.Artifact> artifact
...
+                    graphRoots.stream().map(GraphRootMatcher::new).collect(Collectors.toList());
+
+            DependencyMatcher subTreeMatcher = new OrDependencyMatcher(filterMatchers);
+
+            Set<Artifact> artifacts = new HashSet<>();
+            for (Dependency dep : project.getDependencies()) {
+                if (subTreeMatcher.matches(dep)) {
+                    artifacts.addAll(resolveDependencyArtifacts(dep));
+                }
+            }
```
mthmulders: For consistent code style, did you consider using Streams here as well, just like a few lines above?

src/main/java/org/apache/maven/plugins/dependency/fromDependencies/GraphRoot.java
```diff
@@ -0,0 +1,55 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.plugins.dependency.fromDependencies;
+
+/**
+ * All values are static, no expression, so matcher can use equals
```
mthmulders: But the matcher does not use `equals()`?

src/main/java/org/apache/maven/plugins/dependency/fromDependencies/GraphRootMatcher.java
```diff
@@ -0,0 +1,41 @@
...
+
+    private final GraphRoot graphRoot;
+
+    GraphRootMatcher(GraphRoot graphRoot) {
+        this.graphRoot = graphRoot;
+    }
+
+    @Override
+    public boolean matches(Dependency dependency) {
+        return matches(graphRoot.getGroupId(), dependency.getGroupId())
```
mthmulders: I think you could've used `Objects.equals()` here - it's available since Java 1.7, according to its Javadoc.

src/it/projects/copy-dependencies-graphroots/pom.xml
```diff
@@ -0,0 +1,103 @@
...
+    <version>5</version>
+  </parent>
+
+  <groupId>org.apache.maven.its.dependency</groupId>
+  <artifactId>test</artifactId>
+  <version>1.0-SNAPSHOT</version>
+
+  <name>Test</name>
+  <description>
+    Test dependency:copy-dependencies
```
mthmulders:
```suggestion
    Test dependency:copy-dependencies using graphRoots
```

src/main/java/org/apache/maven/plugins/dependency/fromDependencies/GraphRootMatcher.java
```diff
@@ -0,0 +1,41 @@
...
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package org.apache.maven.plugins.dependency.fromDependencies;
+
+import org.apache.maven.model.Dependency;
+
+class GraphRootMatcher implements DependencyMatcher {
```
mthmulders: Personally, I would've added unit tests for this class.

src/main/java/org/apache/maven/plugins/dependency/fromDependencies/AbstractDependencyFilterMojo.java
```diff
@@ -479,6 +502,38 @@ private Set<Artifact> resolve(Set<org.eclipse.aether.artifact.Artifact> artifact
...
+    }
+
+    private Set<Artifact> resolveDependencyArtifacts(Dependency root) throws DependencyResolutionException {
+        org.eclipse.aether.graph.Dependency dependency = RepositoryUtils.toDependency(
+                root, session.getRepositorySession().getArtifactTypeRegistry());
+
+        List<RemoteRepository> remoteRepositories =
+                RepositoryUtils.toRepos(session.getProjectBuildingRequest().getRemoteRepositories());
+
+        Collection<org.eclipse.aether.artifact.Artifact> depArtifacts =
```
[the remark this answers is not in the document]
rfscholte: The project dependencies have already been resolved, as the goal of the final Mojo contains `requiresDependencyResolution = ResolutionScope.TEST`. So it is actually re-resolving the dependencies. The MavenProject doesn't provide the graph, hence the need ask for the graph user maven artifact resolver.

src/main/java/org/apache/maven/plugins/dependency/fromDependencies/GraphRoot.java
rfscholte: This is a good question. My first approach was to make these values Nullable: if not specified it is a match. However while working with it (and as confirmed in the integration-test) it is probably better to be explicit. Just always specify groupId and artifactId, and type if it is not a jar, just as explicit as the dependency. WDYT?

src/main/java/org/apache/maven/plugins/dependency/fromDependencies/GraphRootMatcher.java
rfscholte: The previous comment. If graphroot is explicit, this matches can be removed.

src/main/java/org/apache/maven/plugins/dependency/fromDependencies/GraphRoot.java
mthmulders: I would certainly prefer to make it explicit, and not give `null` the magical meaning of "could be anything" or "don't care".

> Just always specify groupId and artifactId, and type if it is not a jar, just as explicit as the dependency.

You mean: actual matching is as explicit as the graphRoot declaration in the plugin configuration?

src/main/java/org/apache/maven/plugins/dependency/fromDependencies/GraphRootMatcher.java
rfscholte: I've decided to keep this class, as we're trying to match 2 different kind of classes. null-ignoring matches method has been replaced.

src/main/java/org/apache/maven/plugins/dependency/fromDependencies/AbstractDependencyFilterMojo.java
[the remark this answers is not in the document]
rfscholte: done

mthmulders (approved):

slawekjaranowski (approved):

src/main/java/org/apache/maven/plugins/dependency/fromDependencies/AbstractDependencyFilterMojo.java
```diff
@@ -479,6 +502,42 @@ private Set<Artifact> resolve(Set<org.eclipse.aether.artifact.Artifact> artifact
...
+            return artifacts;
+        }
+    }
+
+    private Set<Artifact> resolveDependencyArtifacts(Dependency root) throws DependencyResolutionException {
+        org.eclipse.aether.graph.Dependency dependency = RepositoryUtils.toDependency(
+                root, session.getRepositorySession().getArtifactTypeRegistry());
+
+        List<RemoteRepository> remoteRepositories =
+                RepositoryUtils.toRepos(session.getProjectBuildingRequest().getRemoteRepositories());
```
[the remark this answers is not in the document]
rfscholte: I wasn't expecting "aether" in MavenProject, but if it is there... why not use it. Done.

PR #1599, opened 2026-04-01
Add dependency:add and dependency:remove goals

brunoborges:
## Summary

This PR adds two new Maven Dependency Plugin goals: `dependency:add` and `dependency:remove`. These goals provide a CLI-based way to manage dependencies without manually editing `pom.xml`.

### Motivation

Most major dependency management ecosystems provide a single-command way to add a dependency:
- **Python:** `pip install google-adk`
- **Node:** `npm install @google/adk`
- **Rust:** `cargo add google-adk`
- **.NET:** `dotnet add package Google.Adk`
- **Maven:** manual XML editing required

This friction is a barrier for developers coming from other ecosystems, and for AI coding agents where a CLI invocation is safer and more deterministic than direct XML manipulation.

#### `dependency:add`

Adds a dependency to the project's `pom.xml`:

```bash
mvn dependency:add -Dgav=org.apache.commons:commons-lang3:3.17.0
mvn dependency:add -Dgav=org.apache.commons:commons-lang3:3.17.0 -Dscope=test
mvn dependency:add -Dgav=org.apache.commons:commons-lang3:3.17.0 -Dmanaged
mvn dependency:add -Dgav=org.springframework.boot:spring-boot-dependencies:pom:3.2.0 -Dscope=import -Dmanaged
```

Features:
- GAV-only coordinate input via `-Dgav`, using `groupId:artifactId[:extension[:classifier]]:version`
- Separate `-Dscope`, `-Dtype`, `-Dclassifier`, and `-Doptional` parameters for dependency attributes
- `-Dmanaged` flag for adding to `<dependencyManagement>`
- BOM imports via explicit `pom` extension plus `scope=import` and `managed`
- `-Dprofile` to target an existing Maven profile
- Maven `-pl` support for targeting child modules in multi-module projects
- Version inference from parent `<dependencyManagement>` when adding a version-less dependency
- Duplicate detection that fails with a descriptive error instead of silently updating existing entries
- Type/classifier-aware matching for duplicate detection
- Cross-reference against Maven's resolved model to catch property-interpolated duplicates
- Parent POM inheritance warning in multi-module projects

#### Convention auto-detection (`-Dalign`, enabled by default)

When adding a dependency, the goal can detect and follow existing dependency management conventions:

- **Managed dependency detection:** if most existing `<dependencies>` entries are version-less, the new dependency is added as a managed dependency in the nearest parent POM and as a version-less dependency in the child
- **Version property detection:** if most existing versions use `${...}` property references, a version property is automatically created
- **Property naming convention:** recognizes `.version`, `-version`, camelCase `Version`, and `version.` prefix patterns

| Parameter | Default | Description |
|-----------|---------|-------------|
| `align` | `true` | Enable convention auto-detection |
| `useProperty` | auto | Force/prevent version property creation |
| `propertyName` | auto | Explicit property name, overriding pattern detection |

Explicit flags such as `managed`, `useProperty`, `propertyName`, and `align=false` override detected conventions.

Example in a multi-module project whose parent uses `-version` suffix properties:

```bash
mvn dependency:add -Dgav=com.google.guava:guava:33.0.0-jre
```

Automatically produces:
- **parent/pom.xml**: `<guava-version>33.0.0-jre</guava-version>` property plus managed dependency with `${guava-version}`
- **child/pom.xml**: version-less `<dependency>` entry

#### `dependency:remove`

Removes a dependency from the project's `pom.xml`:

```bash
mvn dependency:remove -Dgav=org.apache.commons:commons-lang3
mvn dependency:remove -Dgav=org.apache.commons:commons-lang3 -Dmanaged
mvn dependency:remove -Dgav=com.example:lib -Dtype=test-jar
mvn dependency:remove -Dgav=com.example:lib -Dclassifier=sources
```

Features:
- GAV-only coordinate input; only `groupId` and `artifactId` are required for removal
- `-Dmanaged` for removing from `<dependencyManagement>`
- `-Dprofile` to target an existing Maven profile
- Maven `-pl` support for targeting child modules in multi-module projects
- Type/classifier-aware matching for precise removal
- Child module safety warning when removing a managed dependency that child modules reference without an explicit version
- Support for module entries that point directly to a module POM file, such as `<module>my-module/my-pom.xml</module>`

### Implementation Details

- **POM editing:** Uses [DomTrip](https://github.com/maveniverse/domtrip) for formatting-preserving XML edits, including comments, indentation, namespaces, encoding, BOM markers, and XML declarations
- **Security:** Rejects `DOCTYPE`/`ENTITY` declarations before parsing and validates the root element is `<project>`
- **Robustness:** Writes through a temporary file before replacing the target POM and refreshes Maven's in-memory model after edits
- **Goal structure:** `AddDependencyMojo` and `RemoveDependencyMojo` extend `AbstractDependencyMojo`, use constructor injection, and follow the existing `doExecute()` pattern
- **Dependency model:** Adds `DependencyEntry` for parsing Maven dependency entries, including the version-optional `groupId:artifactId` form needed by `dependency:remove`
- **Convention detection:** Uses a `PropertyPattern` enum for type-safe pattern matching and scans child/parent POMs through DomTrip

### Testing

- Added unit coverage for `DependencyEntry`, `AddDependencyMojo`, `RemoveDependencyMojo`, and DomTrip-based POM editing behavior
- Added integration test projects for add basic, add GAV shorthand, add managed, add multi-module, add multi-module with dash-style version properties, remove basic, and remove not-found scenarios
- Site output was checked with `mvn site -P reporting`

### Documentation

- Goals overview lists `dependency:add` and `dependency:remove`
- New site page: `src/site/apt/examples/managing-dependencies.apt.vm`
- Usage page updated with a Managing Dependencies section
- Full Javadoc and `@since 3.11.0` annotations added for the new goals, parameters, and support classes

### Checklist

Following this checklist to help us incorporate your contribution quickly and easily:

- [x] Each commit in the pull request should have a meaningful subject line and body. Note that commits might be squashed by a maintainer on merge.
- [x] Run `mvn verify` to make sure basic checks pass. A more thorough check will be performed on your pull request automatically.

To make clear that you license your contribution under the [Apache License Version 2.0, January 2004](http://www.apache.org/licenses/LICENSE-2.0) you have to acknowledge this by using the following check-box.


src/main/java/org/apache/maven/plugins/dependency/pom/DependencyCoordinates.java
```diff
@@ -0,0 +1,197 @@
...
+ * Represents parsed Maven dependency coordinates (GAV + scope/type/classifier).
+ * Supports parsing from a colon-separated string in the format:
+ * {@code groupId:artifactId[:version[:scope[:type[:classifier]]]]}.
+ *
+ * @since 3.11.0
+ */
+public class DependencyCoordinates {
+
+    private static final Set<String> VALID_SCOPES =
+            new HashSet<>(Arrays.asList("compile", "provided", "runtime", "test", "system", "import"));
```
rmannibucau, line 47: misses a few like `test-runtime`, maybe look to Maven4ScopeManagerConfiguration (or its v3 flavor)

src/main/java/org/apache/maven/plugins/dependency/pom/DependencyEntry.java
```diff
@@ -0,0 +1,197 @@
...
+            throw new IllegalArgumentException("GAV string must not be null or empty");
+        }
+        // Use split with -1 limit to preserve trailing empty segments for validation
+        String[] tokens = gav.split(":", -1);
+        if (tokens.length < 2 || tokens.length > 6) {
+            throw new IllegalArgumentException("Invalid GAV format: '" + gav
+                    + "'. Expected groupId:artifactId[:version[:scope[:type[:classifier]]]]");
+        }
+        DependencyCoordinates coords = new DependencyCoordinates();
+        coords.groupId = tokens[0].trim();
```
rmannibucau, line 90: think order should be aligned  with `org.eclipse.aether.artifact.DefaultArtifact#DefaultArtifact(java.lang.String, java.util.Map<java.lang.String,java.lang.String>, org.eclipse.aether.artifact.ArtifactType)` so we have a single way to define the coordinates so more `<groupId>:<artifactId>[:<extension>[:<classifier>]]:<version>`

additional entries can be added in other properties/parameters

src/main/java/org/apache/maven/plugins/dependency/pom/DependencyEntry.java
```diff
@@ -0,0 +1,197 @@
...
+        }
+        return coords;
+    }
+
+    /**
+     * Validates that required fields are present.
+     *
+     * @throws IllegalArgumentException if groupId or artifactId is missing
+     */
+    public void validate() {
```
rmannibucau, line 137: +version to avoid weirdish side effects?

src/main/java/org/apache/maven/plugins/dependency/pom/DependencyEntry.java
```diff
@@ -0,0 +1,197 @@
...
+ */
+public class DependencyCoordinates {
+
+    private static final Set<String> VALID_SCOPES =
+            new HashSet<>(Arrays.asList("compile", "provided", "runtime", "test", "system", "import"));
+
+    private String groupId;
+    private String artifactId;
+    private String version;
+    private String scope;
```
rmannibucau, line 52: scope and optional are not part of coordinates so might be great to find another name? theoritically type too (extension is) but we can live with this since we already have this inconsistency elsewhere

src/main/java/org/apache/maven/plugins/dependency/pom/PomEditor.java
```diff
@@ -0,0 +1,697 @@
...
+
+            DocumentBuilderFactory factory = DocumentBuilderFactory.newInstance();
+            factory.setNamespaceAware(true);
+            factory.setFeature("http://apache.org/xml/features/nonvalidating/load-external-dtd", false);
+            factory.setFeature("http://xml.org/sax/features/external-general-entities", false);
+            factory.setFeature("http://xml.org/sax/features/external-parameter-entities", false);
+            factory.setFeature(javax.xml.XMLConstants.FEATURE_SECURE_PROCESSING, true);
+            factory.setFeature("http://apache.org/xml/features/disallow-doctype-decl", true);
+            DocumentBuilder builder = factory.newDocumentBuilder();
+            Document doc = builder.parse(pomFile);
```
rmannibucau: why not using maven service to load the pom `Model` instead of using raw xml?

src/main/java/org/apache/maven/plugins/dependency/pom/PomEditor.java
```diff
@@ -0,0 +1,697 @@
...
+        depsElement.removeChild(dep);
+        return true;
+    }
+
+    /**
+     * Writes the modified POM back to disk using an atomic write (temp file + rename).
+     *
+     * @throws IOException if the file cannot be written
+     */
+    public void save() throws IOException {
```
rmannibucau: let's forbid any XML API there since it will break for sure the pom and we don't want that so ideally we just do byte manipulation since we don't even need string level IMHO

src/main/java/org/apache/maven/plugins/dependency/AddDependencyMojo.java
```diff
@@ -0,0 +1,303 @@
...
+     * Optional when adding to {@code <dependencies>} if the version is managed by an ancestor.
+     */
+    @Parameter(property = "version")
+    private String version;
+
+    /**
+     * Shorthand coordinates: {@code groupId:artifactId[:version[:scope[:type[:classifier]]]]}.
+     */
+    @Parameter(property = "gav")
+    private String gav;
```
rmannibucau, line 81: I'd choose between gav or split flavor but I wouldn't start by mixing both, think split flavor is more common but both leads to inconsistencies we must validate for a very low gain

src/main/java/org/apache/maven/plugins/dependency/AddDependencyMojo.java
```diff
@@ -0,0 +1,303 @@
...
+     * Whether the dependency is optional. Setting {@code -Doptional=false} explicitly
+     * removes the {@code <optional>} element when updating an existing dependency.
+     */
+    @Parameter(property = "optional")
+    private Boolean optional;
+
+    /**
+     * When {@code true}, insert into {@code <dependencyManagement>} instead of {@code <dependencies>}.
+     */
+    @Parameter(property = "managed", defaultValue = "false")
```
rmannibucau, line 112: should it be an enum (NO, YES, BOTH - names are bad but you get the idea)

src/main/java/org/apache/maven/plugins/dependency/AddDependencyMojo.java
```diff
@@ -0,0 +1,303 @@
...
+    private boolean managed;
+
+    /**
+     * Target a specific Maven profile by its {@code <id>}. When set, the dependency is added
+     * to the profile's {@code <dependencies>} or {@code <dependencyManagement>} section.
+     * The profile must already exist in the POM.
+     */
+    @Parameter(property = "profile")
+    private String profile;
```
rmannibucau, line 156: can't it be a list? 🤔

src/main/java/org/apache/maven/plugins/dependency/AddDependencyMojo.java
```diff
@@ -0,0 +1,303 @@
...
+
+    /**
+     * When {@code true}, add as a BOM import ({@code type=pom}, {@code scope=import})
+     * into {@code <dependencyManagement>}.
+     */
+    @Parameter(property = "bom", defaultValue = "false")
+    private boolean bom;
```
rmannibucau: not needed?

src/main/java/org/apache/maven/plugins/dependency/AddDependencyMojo.java
```diff
@@ -0,0 +1,303 @@
...
+                editor.setProfileId(profile);
+            }
+            Element existing = editor.findDependency(
+                    coords.getGroupId(),
+                    coords.getArtifactId(),
+                    coords.getType(),
+                    coords.getClassifier(),
+                    targetManaged);
+
+            if (existing != null) {
```
rmannibucau: fail? we are in add goal, not update one so this is very misleading and error prone IMHO

src/main/java/org/apache/maven/plugins/dependency/AddDependencyMojo.java
```diff
@@ -0,0 +1,303 @@
...
+                // Dependency exists in the resolved model but not in raw XML — property interpolation
+                throw new MojoFailureException("Dependency " + coords.getGroupId() + ":" + coords.getArtifactId()
+                        + " already exists in the POM (using property references). "
+                        + "Cannot safely add or update automatically. Please edit the POM manually.");
+            } else {
+                editor.addDependency(coords, targetManaged);
+                getLog().info("Added dependency " + coords + " to " + pomFile.getName());
+            }
+
+            editor.save();
```
rmannibucau: 🤔 also add in model and `MavenProject` so if done in a chain it doesn't break (misses) it?

src/main/java/org/apache/maven/plugins/dependency/AddDependencyMojo.java
```diff
@@ -0,0 +1,303 @@
...
+        }
+
+        try {
+            coords.validate();
+        } catch (IllegalArgumentException e) {
+            throw new MojoFailureException(e.getMessage());
+        }
+
+        // Convert NONE sentinels to empty strings (signals field removal during update)
+        if ("NONE".equals(coords.getScope())) {
```
rmannibucau: why do we use NONE and not "" as default so no code is needed?

src/main/java/org/apache/maven/plugins/dependency/RemoveDependencyMojo.java
```diff
@@ -0,0 +1,250 @@
...
+                            + coords.getArtifactId()
+                            + " exists in " + section + " but uses property references in the POM. "
+                            + "Please remove it manually.");
+                }
+                String section = targetManaged ? "<dependencyManagement>" : "<dependencies>";
+                throw new MojoFailureException("Dependency " + coords.getGroupId() + ":" + coords.getArtifactId()
+                        + " not found in " + section + ".");
+            }
+
+            editor.save();
```
rmannibucau: similar thought to have about aligning the in mem model with file system change

src/main/java/org/apache/maven/plugins/dependency/SearchDependencyMojo.java
```diff
@@ -0,0 +1,675 @@
...
+    /**
+     * Maximum number of results to return.
+     */
+    @Parameter(property = "rows", defaultValue = "10")
+    private int rows;
+
+    /**
+     * Maven Central Search API endpoint.
+     */
+    @Parameter(property = "repositoryUrl", defaultValue = "https://search.maven.org/solrsearch/select")
```
rmannibucau: dont think we want to depend on an external service like that so if there is a search mojo it should rely on resolver API or not be IMHO

src/main/java/org/apache/maven/plugins/dependency/SearchDependencyMojo.java
```diff
@@ -0,0 +1,675 @@
...
+                urlStr += "&core=gav";
+            }
+            URL url = new URL(urlStr);
+
+            connection = (HttpURLConnection) url.openConnection();
+            connection.setRequestMethod("GET");
+            connection.setRequestProperty("Accept", "application/json");
+            connection.setRequestProperty(
+                    "User-Agent", "Apache-Maven-Dependency-Plugin/" + getPluginVersion() + " (dependency:search)");
+            connection.setConnectTimeout(10000);
```
rmannibucau: might be worth configuring - not relevant using aether since it is by default

src/main/java/org/apache/maven/plugins/dependency/SearchDependencyMojo.java
```diff
@@ -0,0 +1,675 @@
...
+        } catch (IOException e) {
+            throw new MojoExecutionException("Failed to query Maven Central search API: " + e.getMessage(), e);
+        } finally {
+            if (connection != null) {
+                connection.disconnect();
+            }
+        }
+    }
+
+    private String readResponse(InputStream inputStream) throws IOException {
```
rmannibucau: ByteArrayOutputStream + new String sounds more relevant and doesn't silently drop characters (even if not critical for json, if you want to read error stream to have a better error message it can be )

src/main/java/org/apache/maven/plugins/dependency/SearchDependencyMojo.java
```diff
@@ -0,0 +1,675 @@
...
+        try (BufferedReader reader = new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8))) {
+            String line;
+            while ((line = reader.readLine()) != null) {
+                sb.append(line);
+            }
+        }
+        return sb.toString();
+    }
+
+    private static String readErrorResponse(HttpURLConnection connection) {
```
rmannibucau: no need to try to be clever and truncate, will likely truncate what has the important information - we failed hard on that in surefire for ex

src/main/java/org/apache/maven/plugins/dependency/SearchDependencyMojo.java
```diff
@@ -0,0 +1,675 @@
...
+                    }
+                }
+            }
+            return sb.toString();
+        } catch (IOException e) {
+            return "";
+        }
+    }
+
+    void displayResults(String jsonResponse) throws MojoExecutionException {
```
rmannibucau: 🤔 

I'd use a single log.info to have a better formatting and aggregation of log depending the environment it runs within

also maybe ensure it can be dumped as json and optionally in a configured file to enable "pipelines" reusing this extraction (list goal is used like that for ex)

src/main/java/org/apache/maven/plugins/dependency/SearchDependencyMojo.java
```diff
@@ -0,0 +1,675 @@
...
+        }
+    }
+
+    // ── Interactive mode ────────────────────────────────────────────────
+
+    /**
+     * Whether interactive mode should be used. Package-private for testing.
+     */
+    boolean isInteractive() {
+        return interactive && System.console() != null;
```
rmannibucau: console can be null in idea console ;)

on java >= 22 we can call `java.io.Console#isTerminal` so either we say it is ok to not support interactive mode < 22 or we do test the common env var or just use the toggle IMHO

src/main/java/org/apache/maven/plugins/dependency/SearchDependencyMojo.java
```diff
@@ -0,0 +1,675 @@
...
+
+    private void interactiveLoop(String initialJson) throws MojoExecutionException, MojoFailureException {
+        String currentJson = initialJson;
+        String currentQuery = query.trim();
+
+        while (true) {
+            List<String[]> artifacts = extractArtifacts(currentJson);
+            int numFound = extractInt(currentJson, "numFound");
+
+            getLog().info("Search results for: " + currentQuery);
```
rmannibucau: not a fan of using logger as a line formatter, this doesn't work at all when you change its formatter so maybe collapse lines - it is fine to put `\n` but a single call per info

src/main/java/org/apache/maven/plugins/dependency/SearchDependencyMojo.java
```diff
@@ -0,0 +1,675 @@
...
+                }
+                currentQuery = input.trim();
+                currentJson = performSearch(currentQuery, rows, false);
+                continue;
+            }
+
+            displayNumberedResults(artifacts, numFound);
+
+            String input = readLine("Enter number to select, text to search again, 'q' to quit: ");
+            if (input == null
```
rmannibucau: you can trim only once i guess

src/main/java/org/apache/maven/plugins/dependency/SearchDependencyMojo.java
```diff
@@ -0,0 +1,675 @@
...
+     * @return the user input, or {@code null} if no console is available
+     */
+    String readLine(String prompt) {
+        if (System.console() == null) {
+            return null;
+        }
+        return System.console().readLine("%s", prompt);
+    }
+
+    private void interactiveLoop(String initialJson) throws MojoExecutionException, MojoFailureException {
```
rmannibucau: im not sure this is needed at all, looks way more complex to use than just rerunning a command, I would drop it for the first version

src/main/java/org/apache/maven/plugins/dependency/SearchDependencyMojo.java
```diff
@@ -0,0 +1,675 @@
...
+        for (int i = 0; i < length; i++) {
+            sb.append('\u2500');
+        }
+        return sb.toString();
+    }
+
+    /**
+     * Extracts artifact info (groupId, artifactId, latestVersion) from the Solr JSON response.
+     */
+    static List<String[]> extractArtifacts(String json) {
```
rmannibucau: we can bring johnzon-core I guess to do the parsing

src/main/java/org/apache/maven/plugins/dependency/SearchDependencyMojo.java
```diff
@@ -0,0 +1,675 @@
...
+        Matcher matcher = pattern.matcher(json);
+        if (matcher.find()) {
+            return Integer.parseInt(matcher.group(1));
+        }
+        return 0;
+    }
+
+    private String getPluginVersion() {
+        Package pkg = getClass().getPackage();
+        if (pkg != null && pkg.getImplementationVersion() != null) {
```
rmannibucau: you have it in plugin.xml, not sure how this one is reliable

src/test/java/org/apache/maven/plugins/dependency/pom/PomEditorTest.java
```diff
@@ -0,0 +1,1005 @@
...
+import org.w3c.dom.Element;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+import static org.junit.jupiter.api.Assertions.assertFalse;
+import static org.junit.jupiter.api.Assertions.assertNotNull;
+import static org.junit.jupiter.api.Assertions.assertNull;
+import static org.junit.jupiter.api.Assertions.assertThrows;
+import static org.junit.jupiter.api.Assertions.assertTrue;
+
+class PomEditorTest {
```
rmannibucau, line 44: neat pick but can be worth adding the string where the `contains` method is called on in all error messages and not just a description (for CI failures)

gnodet (changes requested):
## Review Summary — PR #1599

### What it does
Adds `dependency:add`, `dependency:remove`, and `dependency:search` goals to give Maven a CLI-based dependency management experience like `cargo add` / `npm install`.

### Verdict: Request Changes

---

### Architectural / Design Issues

**1. Should use DomTrip for POM editing**
The PR hand-rolls ~450 lines of DOM manipulation (`PomEditor.java`) using `javax.xml.parsers` + `Transformer`. [DomTrip](https://github.com/maveniverse/domtrip) is purpose-built for this, already in Maven 4 core, and [Toolbox](https://github.com/maveniverse/toolbox) already uses it. The `Transformer` API is unreliable for preserving formatting (the code already has workarounds for `standalone="no"` injection).

**2. Should use Maven Search API (MASE) for search**
The PR hits the old `search.maven.org/solrsearch/select` endpoint with raw `HttpURLConnection` and ~200 lines of hand-rolled JSON parsing. Toolbox uses the MASE library with proper `SearchBackend` abstraction, typed queries, and defaults to the current `central.sonatype.com` endpoint. The hand-rolled JSON parser doesn't handle all escape sequences and can false-match field names inside string values.

**3. Partial overlap with Maveniverse Toolbox**
Toolbox has `add-managed-dependency` and `gav-search` already. However, `add-dependency`, `remove-dependency`, and `remove-managed-dependency` are genuinely new — Toolbox has the internal primitives (`DEPENDENCIES`/`MANAGED_DEPENDENCIES` + `UPSERT`/`DELETE`) but doesn't expose them as mojos yet. The overlap is partial, not total.

**4. GAV format doesn't follow Maven conventions**
The PR uses `g:a[:v[:scope[:type[:classifier]]]]`. Standard Maven coordinate format is `g:a[:extension[:classifier]]:version` (as in `DefaultArtifact`). Scope and optional are not part of coordinates and should only be separate parameters.

---

### Functional Gaps

**5. No version property creation**
Versions are written as literal strings (`<version>1.0.0</version>`). No support for creating/reusing `<properties>` entries — a pattern most real Maven projects follow.

**6. No combo managed + dependency mode**
Can't add a `<dependencyManagement>` entry and a `<dependencies>` reference in one command. No coordinated workflow.

**7. No upgrade/update path**
`add` silently doubles as update. No dedicated update goal, no latest-version resolution, no integration with Maven Versions Plugin.

---

### Complexity for Little Value

**8. Interactive TUI should be dropped for v1**
Uses raw `System.console()` (no JLine), doesn't work in IDEs, adds ~200 lines of code. The non-interactive table output is sufficient.

---

_Claude Code on behalf of Guillaume Nodet_

brunoborges:
Hey @gnodet, you requested changes here, but then you also created a competing PR #1600.

I'm cool with that, because all I want is to be able to do `mvn dependency:add`. Having the credit to an open source contribution is far from my priorities.

So for this reason, I will focus on @rmannibucau's feedback for now, as I see this approach straightforward and not dependant on non-Apache libraries like `domtrip`.

### What has changed in a18c1c5

1. **`test-runtime` added** to `VALID_SCOPES`
2. **GAV format aligned** to `g:a[:ext[:cls]]:v` (matching `DefaultArtifact`). Scope removed from GAV — must be passed via `-Dscope`.
3. **Duplicate dependency now fails** instead of auto-updating (add ≠ update)
4. **In-memory model sync** after `PomEditor.save()` in both Add and Remove mojos
5. **NONE sentinel removed** — empty string is used instead
6. **`ByteArrayOutputStream`** for HTTP response reading
7. **No error message truncation**
8. **Single `log.info` calls** with `\n` instead of multiple calls
9. **Interactive TUI dropped entirely** (~400 lines of tests, ~200 lines of source removed)
10. **`PluginDescriptor`** for version detection (null-safe)
11. **Test assertion messages** include the actual string being checked

### Open questions — need your guidance

I'd appreciate your thoughts on these design decisions before implementing:

1. **PomEditor approach** — You mentioned byte manipulation (L314) and also asked about Maven service (L104). For the first version, should we keep the current DOM approach and plan byte manipulation / Maven service for a follow-up, or is this a blocker?

2. **GAV vs split parameters** (L86) — You suggested picking one. Which do you prefer? I'm leaning toward keeping both since GAV is convenient for CLI use (`-Dgav=g:a:v`) while split params are better for POM-based plugin configuration. Happy to drop one if you feel strongly.

3. **`managed` as enum** (L119) — `YES/NO/BOTH` — is this about allowing the user to add a dependency to *both* `<dependencies>` and `<dependencyManagement>` in one invocation?

4. **`gav` as list** (L128) — For batch operations like `-Dgav=g1:a1:v1,g2:a2:v2`? Any preference on the delimiter?

5. **`bom` flag** (L135) — You said "not needed?" — should I remove it entirely and let users just set `-Dtype=pom -Dscope=import -Dmanaged=true` explicitly?

6. **Resolver API vs external search** (L67) — Fully agree this would be better. Is there a specific resolver API you have in mind for artifact search/discovery? The current Maven resolver APIs I'm aware of don't have a search/query capability equivalent to Central Search.

7. **johnzon-core for JSON parsing** (L496) — Happy to add this dependency. Should I replace all the manual JSON parsing with it? 

8. **domtrip for XML manipulation** — Are we OK with using a non-Apache 3rd-party dependency? 

9. **Rename `DependencyCoordinates`** (L40) — Since it holds scope/optional which aren't coordinates, any name suggestion? `DependencyDescriptor`? `DependencySpec`?

10. **Add version to `validate()`** (L98) — Should `validate()` require version to be non-null/non-empty, or just warn if missing?

src/main/java/org/apache/maven/plugins/dependency/pom/DependencyCoordinates.java
```diff
@@ -25,14 +25,19 @@
...
+ * {@code org.eclipse.aether.artifact.DefaultArtifact}. Scope and optional
+ * are not part of coordinates and must be specified as separate parameters.</p>
  *
  * @since 3.11.0
  */
 public class DependencyCoordinates {
 
     private static final Set<String> VALID_SCOPES =
-            new HashSet<>(Arrays.asList("compile", "provided", "runtime", "test", "system", "import"));
+            new HashSet<>(Arrays.asList("compile", "provided", "runtime", "test", "system", "import", "test-runtime"));
```
rmannibucau: hmm, no more sure of my comment was clear but we are still incomplete no? thinking to `test-only`, `compile-only` for e
if i didnt give a code pointer `org.apache.maven.api.DependencyScope` or `Maven4ScopeManagerConfiguration` should do it

some are maven 4.1.0 (modelversion) only, not sure how we do want to handle that (just open it or validate it) 🤔

brunoborges: @rmannibucau  I'd like to focus these changes on Maven 3.x and then forward port to the Maven 4.x branch. It wasn't clear to me that `test-only` was a Maven 4 scope. I think we should remove it. See commit [9903dba](https://github.com/apache/maven-dependency-plugin/pull/1599/commits/9903dbaa8cf97401b399359b0a75de10b425a8cc)

Once we land these new goals on Maven Dependency Plugin 3.x, we can look at forward porting to Maven 4.x and add the new scopes.

Thoughts?

rmannibucau: v3 is used with maven 4 so not sure this strategy is that relevant in practise
the v4 branch is to migrate to v4 api, not for maven 4 compatibility strictly speaking

that said some reflection can make it since dependency scope enum is available on v4 and quite easy to make it match IMHO

brunoborges: I'd rather keep this one simple. I think supporting objective phases is good enough. If a user wants to select a more specific dependency, they should manually modify `pom.xml` after it is added.

Once we land these goals, we can look at additional enhancements for Maven 4. Majority of users (like... _majority_) are on Maven 3.x

gnodet:
> Hey @gnodet, you requested changes here, but then you also created a competing PR #1600.
> 
> I'm cool with that, because all I want is to be able to do `mvn dependency:add`. Having the credit to an open source contribution is far from my priorities.

Yes, I tasked an AI agent to see what it would give with the changes I wanted, and got caught in the process...  and I really don't care about the credit either...

A major point, which I think is really important, is the ability to actually use those new goals on real projects, not just starter projects with a very simple structure.  So we need to support adding agents and kinda follow the existing organisation/layout of the project.  Which means, not only simply adding a dependency, but adding it the way other dependencies are expressed: this includes adding a managed dependency in a top level parent and/or using a property.  That's already implemented in my alternative PR.

> So for this reason, I will focus on @rmannibucau's feedback for now, as I see this approach straightforward and not dependant on non-Apache libraries like `domtrip`.

Given 99% of DomTrip code has been written by Maven committers, directly for Maven or related projects, and given the fact it's already a dependency of Maven 4, I don't really think that's a problem.

rmannibucau:
@gnodet was domtrip fixed to keep null/empty tags (both cases) and xsd declaration formatting? At some point I got bitten by these 2/3 cases but was in early stages.

@brunoborges while domtrip can be discussed cause kind of part of maven community as Guillaume mentionned, your PomEditor has issues on the formatting and creates yet another way to rewrite the pom so from my window it is either plain string manipulation specific to the dependency case like IDE do or some investment in domtrip as initially mentionned on the list before the PR and in previous review cycle

guess it can also be saner to drop search mojo from a pr with add/remove (add remove are mainly some internal decisions about what we want to provide nothing crazy and we can agree quickly I guess but search has this external service dependency which is generally not working outside central so defeats the goal of the mojo which should be able to use company specific  repositories and resolver like features IMHO, ultimately, in current form it can better match sonatype plugin than a maven one to be consistent, that said we have the incremental inde feature which is a way to provide such a feature even if not disk friendly, this is why we should discuss to either go nexus 2/3, artifactory etc or nothing).

brunoborges:
If using `domtrip` is not an issue, let's go with that.

And I agree with the suggestion to drop search.

brunoborges:
## Update: Two major changes pushed

### 1. PomEditor rewritten with DomTrip ([faedbd1d](https://github.com/apache/maven-dependency-plugin/pull/1599/commits/faedbd1d8b04109c95cd9e287104f8480d84a4c6))

Replaced ~450 lines of hand-rolled JAXP/DOM/Transformer code with [DomTrip](https://github.com/maveniverse/domtrip) (`eu.maveniverse.maven.domtrip:domtrip-maven:1.0.0`). PomEditor went from 697 → 343 lines (51% reduction).

DomTrip handles all formatting preservation automatically — BOM, encoding, indentation, line endings, namespaces, and comments. Removed all manual workarounds (standalone="no" cleanup, BOM re-injection, indent detection via GCD, etc.). Kept a pre-parse DOCTYPE rejection check for XXE security.

### 2. Search goal removed ([13152c51](https://github.com/apache/maven-dependency-plugin/pull/1599/commits/13152c51544c4c7c9e738921ebf8cf6d67d44290))

Dropped `dependency:search` entirely (1,032 lines removed). The PR now focuses exclusively on `dependency:add` and `dependency:remove`.

---

### Remaining open questions for @rmannibucau

From the original review feedback, these design decisions still need your input:

1. **GAV vs split parameters** — Keep both (`-Dgav=g:a:v` and `-DgroupId=... -DartifactId=...`) or pick one only?
2. **`managed` as enum** — Change from boolean to `YES/NO/BOTH` to allow adding to both `<dependencies>` and `<dependencyManagement>` in one invocation?
3. **`gav` as list** — Support batch operations like `-Dgav=g1:a1:v1,g2:a2:v2`?
4. **`bom` flag** — Remove it and let users set `-Dtype=pom -Dscope=import -Dmanaged=true` explicitly?
5. **Rename `DependencyCoordinates`** — Since it holds scope/optional which are not coordinates. Suggestions: `DependencyDescriptor`? `DependencySpec`?
6. **Add version to `validate()`** — Should validation require version to be present?
7. **JSON output option** — You mentioned enabling JSON dump and file output for pipeline use. Should we add `-Dformat=json` or `-DoutputFile=...` for the add/remove goals?

brunoborges:
## Latest changes (2 new commits)

### 1. Use `-Dgav` only, remove split params (`4ae00111`)
Removed separate `@Parameter` fields for `groupId`, `artifactId`, and `version` from both `AddDependencyMojo` and `RemoveDependencyMojo`. The `-Dgav` parameter is now the **only** way to specify dependency coordinates.

`-Dscope`, `-Dtype`, `-Dclassifier`, and `-Doptional` remain as separate override parameters (they augment the GAV, not replace it).

This addresses the feedback: *"I'd choose between gav or split flavor but I wouldn't start by mixing both"*.

### 2. Rename `DependencyCoordinates` → `DependencyEntry` (`116a4562`)
Addresses the feedback: *"scope and optional are not part of coordinates so might be great to find another name"*.

---

### Remaining open design questions

1. **`managed` as enum** — Should `-Dmanaged` be an enum (`YES`/`NO`/`BOTH`) instead of a boolean? Currently it's a boolean that toggles between `<dependencies>` and `<dependencyManagement>`.
2. **`bom` flag** — Should `-Dbom` be removed? Currently it's a convenience flag that sets `scope=import` and `type=pom`.
3. **Version in `validate()`** — Should we require version in validation to avoid "weirdish side effects"?
4. **JSON output** — Should the goals support JSON output for pipeline use?

src/main/java/org/apache/maven/plugins/dependency/pom/DependencyEntry.java
brunoborges: We decided to keep version out of `validate()`. The reason is that version requirements are context-dependent:

- `dependency:remove` legitimately needs no version — it matches by `g:a` only
- `dependency:add` to `<dependencyManagement>` requires version (fails explicitly)
- `dependency:add` to `<dependencies>` can infer version from parent dependency management (fails if not found)

Validating version in the value object would either break remove or require separate validation paths. The current approach validates version at the mojo layer where the context is known, and fails with clear error messages in all cases where version is missing and cannot be inferred.

brunoborges:
## Summary of all changes addressing @rmannibucau review

All design questions from the review have been resolved. Here is the full list of commits:

1. **Address review feedback** (`a18c1c5c`) — 14 actionable items (GAV format aligned to Aether, fail on duplicate, model sync, NONE sentinel removed, BAOS for HTTP, PluginDescriptor for version, test assertion messages)
2. **Remove test-runtime scope** (`9903dbaa`) — Focus on Maven 3.x scopes
3. **Replace hand-rolled DOM with DomTrip** (`faedbd1d`) — PomEditor rewritten using DomTrip (697→343 lines)
4. **Remove dependency:search goal** (`13152c51`) — PR now focused on add and remove only
5. **Use -Dgav only** (`4ae00111`) — Removed split groupId/artifactId/version params
6. **Rename DependencyCoordinates → DependencyEntry** (`116a4562`) — Reflects that the class holds more than just coordinates
7. **Fix error messages** (`20433035`) — References -Dgav instead of removed -Dversion
8. **Remove -Dbom flag** (`9c304854`) — Users use explicit `-Dgav=g:a:pom:v -Dscope=import -Dmanaged` instead

### Design decisions made:
- **GAV vs split params**: GAV-only (`-Dgav`)
- **-Dmanaged**: Kept as boolean (two invocations for BOTH case)
- **-Dbom**: Removed (not needed, explicit params suffice)
- **Version in validate()**: Not added — version requirements are context-dependent (remove needs none, add infers from parent)
- **JSON output**: Not applicable to mutation goals (add/remove)
- **List support for gav**: No (single dependency per invocation)

brunoborges:
### Design note: why `DependencyEntry` instead of existing classes

We evaluated replacing `DependencyEntry` with classes already on the classpath:

| Class | GAV parsing | `g:a` (no version) | scope/optional | Verdict |
|---|---|---|---|---|
| `DefaultArtifact` (Aether) | ✅ `g:a[:ext[:cls]]:v` | ❌ Throws | ❌ | Cannot replace parser |
| `Dependency` (maven-model) | ❌ No parser | N/A | ✅ | Cannot replace parser |
| `DependencyEntry` (ours) | ✅ | ✅ | ✅ | Keeps both |

**Key constraint:** `dependency:remove` legitimately accepts `g:a` without a version (it matches by groupId:artifactId only). `DefaultArtifact` throws `IllegalArgumentException` on 2-token inputs because its format requires `<groupId>:<artifactId>[:<extension>[:<classifier>]]:<version>` — version is mandatory.

Using `DefaultArtifact` would require a fallback wrapper for the 2-token case, and using `Dependency` as a carrier would scatter scope validation and `toString()` logic across both mojos. The net effect would be ~80 fewer lines but worse cohesion.

**Decision:** Keep `DependencyEntry` as a focused value object (~228 lines) that bridges GAV parsing + scope validation + all dependency fields in one place.

brunoborges:
### Bug fix: model sync in `dependency:remove` (`d3e86f2`, `e5e6a86`)

Found and fixed a bug in the in-memory model sync during review. When a POM has multiple variants of the same dependency (e.g., both `jar` and `test-jar`), the XML removal correctly targets only the specified variant, but the in-memory model sync was using only `groupId:artifactId` — removing **all** variants from the model. This would cause chained goals to see incorrect state.

**Fix:** Model sync now matches on `groupId:artifactId:type:classifier`, consistent with the XML removal logic.

**Test added:** `modelSyncPreservesOtherVariantsOnRemove` — sets up a POM with both `jar` and `test-jar` variants, removes only `test-jar`, and verifies the `jar` variant remains in the in-memory model.

Also documented type/classifier matching behavior in `managing-dependencies.apt.vm` — explaining that Maven treats `g:a` with different types/classifiers as distinct dependencies.

brunoborges:
## Audit fixes (8a5e16cb)

Performed a thorough audit of all 5 source files (`AddDependencyMojo`, `RemoveDependencyMojo`, `PomEditor`, `DependencyEntry`, `AbstractDependencyMojo`) for NPE risks, validation gaps, security issues, resource leaks, and dead code.

### Fixed in this commit:

1. **🔴 Temp file leak in `PomEditor.save()`** — The old `catch (IOException)` block only cleaned up the temp file on `IOException`. If `toXml()` threw a `RuntimeException`, the temp file would leak. Fixed by using a `finally` block with a success flag to ensure cleanup on any exception path.

2. **🔴 Stale javadoc in `AddDependencyMojo`** — Class javadoc still referenced "BOM import shorthand" which was removed in 9c304854. Removed the stale text.

3. **🟡 Case-sensitive XXE check in `PomEditor.load()`** — The DOCTYPE check used `content.contains("<!DOCTYPE")` which is case-sensitive. While standard XML parsers reject lowercase variants, as defense-in-depth we now use case-insensitive matching and also check for `<!ENTITY` declarations. Added a test for the lowercase bypass.

4. **🟡 Dead code: `updateDependency()` + helpers** — `updateDependency()`, `setOrRemoveChild()`, and `removeChildElement()` had no mojo callers since the `updateExisting` behavior was removed. Removed the dead methods and their 4 tests (net −164 lines).

### Reviewed and accepted (no fix needed):

5. **`threadSafe=true` without synchronization** — Maven's parallel build does not execute the same mojo against the same project instance concurrently, so the model mutation after `save()` is safe in practice.

6. **Path traversal in `checkChildModuleDependencies`** — Module names come from Maven's already-parsed `Model`, which validates module paths during project loading. The code only reads `pom.xml` files. Very low risk.

All 389 tests pass.

brunoborges:
## Second audit fixes (7cbd853e)

Ran a second full audit of all add/remove source files and documentation after the first round of fixes.

### Fixed:

1. **🔴 Stale documentation: "Updating an Existing Dependency" section** — The entire section (lines 95-112 in `managing-dependencies.apt.vm`) described nonexistent behavior: update-on-duplicate semantics and a `NONE` sentinel for clearing fields. In reality, `dependency:add` *fails* on duplicates and no NONE sentinel exists. Replaced with accurate "Duplicate Detection" and "Optional Dependencies" sections.

2. **🔴 Empty-string type/classifier breaks model sync** — When Maven sets unspecified `-Dtype` or `-Dclassifier` to empty string (rather than null), the model sync in `RemoveDependencyMojo` and `existsInResolvedModel` in `AbstractDependencyMojo` would fail to match dependencies with default type `"jar"` because `"".equals("jar")` is false. Fixed by normalizing empty strings to their default values (`"jar"` for type, `""` for classifier).

3. **🟡 NPE in `PomEditor.findProfile(null)`** — `findProfile()` is a public API method that would throw NPE if called with null. Current callers guard against this, but added a null guard for defensive correctness.

4. **🟢 `File.delete()` return value ignored** — `PomEditor.save()` used `tempFile.delete()` whose return value was silently ignored. Replaced with `Files.deleteIfExists()` which throws on failure.

5. **🟢 `-Doptional` parameter undocumented** — Added documentation for the `optional` parameter in `managing-dependencies.apt.vm`.

All 389 tests pass.

brunoborges:
## Third audit fixes (5b43f82b)

### Fixed:

1. **🔴 Stale docs in `usage.apt.vm`** — Line 721 still claimed `dependency:add` "updates automatically" on duplicate. Fixed to say it fails with a descriptive error. Also removed inconsistent quotes from GAV examples to match `managing-dependencies.apt.vm` style.

2. **🟡 Null check for `project.getFile()`** — Both `AddDependencyMojo` and `RemoveDependencyMojo` dereferenced `targetProject.getFile()` without null check. While `requiresProject=true` ensures a project exists, it does not guarantee a physical POM file. Added null checks with clear `MojoExecutionException` messages.

3. **🟡 Null check for `parentProject.getBasedir()`** — `checkChildModuleDependencies` used `getBasedir()` without null check when constructing child module paths. Added guard that skips the check gracefully with a debug log.

4. **🟡 Null validation in `PomEditor.findDependency()`** — Public API method would NPE on null `groupId`/`artifactId` parameters. Added `IllegalArgumentException` with clear message.

All 389 tests pass.

brunoborges:
## Real-world integration testing

Installed the plugin locally (`3.10.1-SNAPSHOT`) and ran **30 tests** across 4 major open-source Maven projects to verify `dependency:add` and `dependency:remove` work correctly in the wild.

### Projects tested

| Project | Description | POM complexity |
|---------|-------------|---------------|
| [Spring Petclinic](https://github.com/spring-projects/spring-petclinic) | Simple Spring Boot app | Standard single-module |
| [Apache Commons Lang](https://github.com/apache/commons-lang) | Apache library | Single-module with managed deps |
| [Apache Camel](https://github.com/apache/camel) | Large integration framework | **4500-line root POM**, 539 property interpolations, massive multi-module |
| [Baeldung Tutorials](https://github.com/eugenp/tutorials) | Hundreds of submodules | Large multi-module |

### Results: 30/30 pass (1 cosmetic observation)

| Test Scenario | Petclinic | Commons Lang | Camel | Tutorials |
|---------------|-----------|-------------|-------|-----------|
| Basic add | ✅ | ✅ | ✅ | ✅ |
| Add with scope | ✅ | ✅ | — | — |
| Add to `<dependencyManagement>` | ✅ | ✅ | ✅ | ✅ |
| Add with `-Doptional=true` | ✅ | — | — | — |
| Add with `-Dtype` / `-Dclassifier` | ✅ | — | — | — |
| Add BOM import (pom + import + managed) | — | — | — | ✅ |
| Version inference from managed deps | — | ✅ | — | — |
| Remove basic | ✅ | — | — | — |
| Remove from `<dependencyManagement>` | — | ✅ | ✅ | — |
| Add then remove round-trip | ✅ | — | ✅ | ✅ |
| Add to submodule | — | — | ✅ | ✅ |
| Duplicate detection (should fail) | ✅ | — | — | — |
| Remove nonexistent (should fail) | ✅ | — | — | — |
| Invalid GAV (should fail) | ✅ | — | — | — |
| Invalid scope (should fail) | — | ✅ | — | — |
| Missing `-Dgav` (should fail) | — | ✅ | — | — |
| Formatting preservation | ✅ | ✅ | ✅ | ✅ |
| POM with property interpolations | — | — | ✅ | — |

### Key findings

- **Formatting preservation is excellent** — round-trip (add then remove) produces zero diff on all projects, including Camel's 4500-line POM with 539 property references.
- **All error cases handled gracefully** — duplicate detection, nonexistent removal, invalid GAV, invalid scope, and missing parameters all produce clear, actionable error messages.
- **Version inference works** — adding without version when a managed version exists correctly omits the `<version>` element.
- **BOM import pattern works** — `dependency:add -Dgav=...dependencies:pom:3.2.0 -Dscope=import -Dmanaged` correctly produces `<type>pom</type>` + `<scope>import</scope>` inside `<dependencyManagement>`.
- **Submodule targeting works** — using `-f path/to/submodule/pom.xml` or running from the submodule directory both work correctly.

### One cosmetic observation

When `dependency:add` creates a new `<dependencies>` section (because one didn't exist) and then `dependency:remove` removes the last entry, the empty container element (`<dependencies/>`) is left behind. This is valid XML and cosmetic only — not a bug.

src/main/java/org/apache/maven/plugins/dependency/pom/DependencyEntry.java
rmannibucau: works for me, just expected a more centralized location with all the validation (and probably an aggregated message with all errors) but not an issue at all 👍

rmannibucau (approved):

gnodet:
> @gnodet was domtrip fixed to keep null/empty tags (both cases) and xsd declaration formatting? At some point I got bitten by these 2/3 cases but was in early stages.

I'm not aware of anything that DomTrip would lost at the moment.  AFAIK, as long as the XML is syntactically parseable, everything is fully preserved.

brunoborges:
How are we feeling about this now?

gnodet:
> How are we feeling about this now?

@brunoborges Could you have a look at https://github.com/brunoborges/maven-dependency-plugin/pull/2/changes ?
The goal is to detect managed dep patterns, version property usage, and naming conventions when adding a property in complex scenarios.

brunoborges:
PR description has been updated to reflect the latest state of the proposed implementation, including the new convention auto-detection feature (`-Dalign`), updated test counts (458+ unit tests, 8 ITs), and implementation details.

brunoborges:
@gnodet and @rmannibucau PR up to date with all feedback given.

brunoborges:
@rmannibucau time to merge?

brunoborges:
> @rmannibucau time to merge?

Wdyt pushing this through this week?

rmannibucau:
@gnodet @cstamas what's the plan with maveniverse work overlapping with that?

@brunoborges is the CI report outdated or did you miss it? :thinking:

brunoborges:
I missed it! Fixing tonight. Update coming tomorrow!

brunoborges:
Updated, @rmannibucau

rmannibucau:
@gnodet @cstamas looks good to me but I'd like your feedback on the concurrent mojo you did before we integrate that

cstamas:
If you mean Pilot, it is not overlapping IMHO. Or wdym?

rmannibucau:
@cstamas a mix between pilot and https://github.com/apache/maven-dependency-plugin/pull/1600, all good for me if we consider this one superseeds it

brunoborges:
Hey all, any update?

src/it/projects/add-dependency/basic/verify.groovy
```diff
@@ -0,0 +1,25 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+File pom = new File(basedir, "pom.xml")
+assert pom.exists()
+def content = pom.text
+assert content.contains('<groupId>org.apache.commons</groupId>')
+assert content.contains('<artifactId>commons-lang3</artifactId>')
+assert content.contains('<version>2.10.1</version>')
```
slawekjaranowski: We have as param: `...add -Dgav=org.apache.commons:commons-lang3:3.17.0`
So why check `2.10.1`

slawekjaranowski: We can use `XmlSlurper` for assertions,

https://docs.groovy-lang.org/latest/html/gapi/groovy/xml/XmlSlurper.html

brunoborges: Thanks @slawekjaranowski ... I addressed these in commit [6737d63](https://github.com/apache/maven-dependency-plugin/pull/1599/commits/6737d6308f44528f0943e3e2981ca12e903f5b46)

brunoborges:
Hey  @slawekjaranowski @rmannibucau @gnodet ...

Are we good with pushing this through?

src/main/java/org/apache/maven/plugins/dependency/resolvers/ResolvePluginsMojo.java
```diff
@@ -153,7 +153,7 @@ public ResolvePluginsMojo(
     @Override
     protected void doExecute() throws MojoExecutionException {
         try {
-            // ideally this should either be DependencyCoordinates or DependencyNode
+            // ideally this should either be DependencyEntry or DependencyNode
```
slawekjaranowski: ?

src/main/java/org/apache/maven/plugins/dependency/AbstractDependencyMojo.java
```diff
@@ -148,4 +153,38 @@ public void setSilent(boolean silent) {
...
         }
     }
+
+    /**
+     * Checks whether the dependency exists in the project's declared (original) model
+     * after property interpolation, but before inheritance merging.
+     * This catches dependencies declared with property references like {@code ${project.groupId}}
+     * without false-positiving on inherited dependencies from a parent POM.
+     */
+    protected static boolean existsInResolvedModel(MavenProject project, DependencyEntry coords, boolean managed) {
```
slawekjaranowski, line 163: I would not like to add a static method to abstract class when it is only used in two specific child .... 
I don't have a good idea for it ... next abstract in level or util class ...

If no better idea can be as is.

brunoborges: I'd prefer to leave it as is, and refactor/change in the future if there is a stronger need for that.

src/main/java/org/apache/maven/plugins/dependency/resolvers/ResolvePluginsMojo.java
brunoborges: Fixed in [0e70709](https://github.com/apache/maven-dependency-plugin/pull/1599/commits/0e70709e37485ece4bc44f010c337cc9d1de9563)

src/main/java/org/apache/maven/plugins/dependency/RemoveDependencyMojo.java
```diff
@@ -0,0 +1,304 @@
...
+        if (parentProject.getBasedir() == null) {
+            getLog().debug("Parent project basedir is null, skipping child module dependency check");
+            return;
+        }
+        StringBuilder affected = new StringBuilder();
+        MavenXpp3Reader pomReader = new MavenXpp3Reader();
+
+        for (String moduleName : parentProject.getModules()) {
+            File moduleDir = new File(parentProject.getBasedir(), moduleName);
+            File modulePom = new File(moduleDir, "pom.xml");
```
slawekjaranowski: we can have in project:

```xml
<module>my-module/my-pom.xml</module>
```

src/main/java/org/apache/maven/plugins/dependency/RemoveDependencyMojo.java
```diff
@@ -0,0 +1,304 @@
...
+
+    /**
+     * Dependency coordinates: {@code groupId:artifactId[:version]}
+     * or {@code groupId:artifactId[:extension[:classifier]]:version}.
+     * Only groupId and artifactId are required. Type and classifier, if provided,
+     * are used for precise matching when multiple dependency variants exist
+     * (e.g., jar vs test-jar).
+     */
+    @Parameter(property = "gav")
+    private String gav;
```
slawekjaranowski, line 77: Please add `@since` tag to all plugin parmeter

src/main/java/org/apache/maven/plugins/dependency/pom/DependencyEntry.java
```diff
@@ -0,0 +1,235 @@
...
+ * {@code Dependency} (maven-model) has all fields but no coordinate string parser.
+ * This class bridges both: it parses {@code g:a} (needed by {@code dependency:remove}),
+ * carries scope/optional, and validates scope against Maven's known values.</p>
+ *
+ * @since 3.11.0
+ */
+public class DependencyEntry {
+
+    private static final Set<String> VALID_SCOPES =
+            new HashSet<>(Arrays.asList("compile", "provided", "runtime", "test", "system", "import"));
```
slawekjaranowski: Maven 4 introduces more scopes, also extension can register own scope ....

I would like to not check it at all - left to user decision and responsibility.

slawekjaranowski (changes requested):
I hop last comments ... 😄

src/main/java/org/apache/maven/plugins/dependency/pom/DependencyEntry.java
brunoborges: Fixed in 7baac9e6

src/main/java/org/apache/maven/plugins/dependency/RemoveDependencyMojo.java
brunoborges: Fixed in 7defb38a

src/main/java/org/apache/maven/plugins/dependency/RemoveDependencyMojo.java
brunoborges: Fixed in 88af9816

brunoborges:
@slawekjaranowski mind to review again?

brunoborges:
Hey @slawekjaranowski @rmannibucau @gnodet ...

Are we good with pushing this through?

src/main/java/org/apache/maven/plugins/dependency/AddDependencyMojo.java
```diff
@@ -0,0 +1,803 @@
...
+ */
+@Mojo(name = "add", requiresProject = true, threadSafe = true)
+public class AddDependencyMojo extends AbstractDependencyMojo {
+
+    /**
+     * Dependency coordinates: {@code groupId:artifactId[:version]}
+     * or {@code groupId:artifactId[:extension[:classifier]]:version}.
+     * Scope, type, classifier, and optional can be overridden via separate parameters.
+     */
+    @Parameter(property = "gav")
```
slawekjaranowski, line 82: + since 😄

slawekjaranowski: and rest of `@Parameter`

src/main/java/org/apache/maven/plugins/dependency/RemoveDependencyMojo.java
```diff
@@ -0,0 +1,321 @@
...
+ */
+@Mojo(name = "remove", requiresProject = true, threadSafe = true)
+public class RemoveDependencyMojo extends AbstractDependencyMojo {
+
+    /**
+     * Dependency coordinates: {@code groupId:artifactId[:version]}
+     * or {@code groupId:artifactId[:extension[:classifier]]:version}.
+     * Only groupId and artifactId are required. Type and classifier, if provided,
+     * are used for precise matching when multiple dependency variants exist
+     * (e.g., jar vs test-jar).
```
slawekjaranowski, line 72: It will be generate in one line in documentation - to check how it will look.

slawekjaranowski:
I'm ok for it ... currently I will release maven-dependency-analyzer naext can be maven-dependency (next week I hope)

We need check site output `mvn site -P reporting`

slawekjaranowski:
@brunoborges please consider to squash to one final commit, or will be sqush during merge

src/main/java/org/apache/maven/plugins/dependency/RemoveDependencyMojo.java
brunoborges: image

Looks fine by me.

brunoborges:
@slawekjaranowski all done!

slawekjaranowski:
@gnodet does your requirements are resolved?

For me looks ok.

cstamas:
LGTM, but unsure is this latest domtrip version?

cstamas (approved):

brunoborges:
> LGTM, but unsure is this latest domtrip version?

Updated in commit 0d65be6

slachiewicz (approved):

slawekjaranowski (approved):

PR #1606, opened 2026-05-12
Remove unused jansi dependency

slawekjaranowski:

- [ ] Your pull request should address just one issue, without pulling in other changes.
- [ ] Run `mvn verify` to make sure basic checks pass.
- [ ] You have run the integration tests successfully (`mvn -Prun-its verify`).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -340,12 +340,6 @@ under the License.
       <scope>test</scope>
     </dependency>
     <dependency>
-      <groupId>org.fusesource.jansi</groupId>
-      <artifactId>jansi</artifactId>
-      <version>2.4.3</version>
-      <scope>test</scope>
-    </dependency>
-    <dependency>
       <groupId>org.assertj</groupId>
       <artifactId>assertj-core</artifactId>
       <version>3.27.7</version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -430,11 +424,9 @@ under the License.
                 <failOnWarning>true</failOnWarning>
                 <ignoredUnusedDeclaredDependencies combine.children="append">
                   <!-- used in tests runtime -->
-                  <ignoredUnusedDeclaredDependency>org.fusesource.jansi:jansi</ignoredUnusedDeclaredDependency>
```
elharo, line 433: How sure are you about this one? It feels risky. Haven't looked at the git log but someone — might even be me — put this here for a reason and even left a comment above as to why. 

If it's genuinely unused now, then remove the comment too.

elharo (changes requested):

pom.xml
slawekjaranowski: comment is valid for rest of ignored dependencies which are still used

elharo: Hmm,

[INFO] --- dependency:3.10.0:analyze (default-cli) @ maven-dependency-plugin ---
[WARNING] Unused declared dependencies found:
[WARNING]    org.apache.maven.resolver:maven-resolver-connector-basic:jar:1.9.25:test
[WARNING]    org.apache.maven.resolver:maven-resolver-transport-file:jar:1.9.25:test
[WARNING]    org.apache.maven.resolver:maven-resolver-transport-http:jar:1.9.25:test


so I guess it's OK to get rid of jansi, but why are these next three still being reported?

elharo (approved):

pom.xml
slawekjaranowski: ok, configuration is in pluginManagment in execution ... try with:

```
mvn dependency:analyze@analyze
```

slawekjaranowski: I will move do build/plugin but in next PR

slawekjaranowski: we have defined base configuration in parent also in execution section ... so move it her to plugin/configuration will lost items from parent

PR #1624, opened 2026-05-17
Log unpacking at debug level

### LINKED ISSUE #1618: Fix inverted logging logic in UnpackUtil.logUnpack()
## Problem

The `logUnpack()` method in `UnpackUtil.java:159-161` has inverted logic:

```java
private void logUnpack(Log logger, File file, File location, String includes, String excludes) {
    if (logger.isInfoEnabled()) {
        return;  // BUG: returns when logging IS enabled
    }
    // ... builds message ...
    logger.info(msg.toString());  // never reached when logging is enabled
}
```

The method returns early when `isInfoEnabled()` is `true`, so log messages are never actually logged.

## Fix

Change the condition to negate the check:

```java
if (!logger.isInfoEnabled()) {
    return;
}
```

## File

`src/main/java/org/apache/maven/plugins/dependency/utils/UnpackUtil.java:158-186`

elharo:
We've been logging this when info logs are not enabled and not logging it when they are. Either way, these logs are noisy so set this to to log at debug level. 

fixes #1618

### DIFF
--- a/src/main/java/org/apache/maven/plugins/dependency/utils/UnpackUtil.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/utils/UnpackUtil.java
@@ -156,32 +156,30 @@ public class UnpackUtil {
     }
 
     private void logUnpack(Log logger, File file, File location, String includes, String excludes) {
-        if (logger.isInfoEnabled()) {
-            return;
+        if (logger.isDebugEnabled()) {
+            StringBuilder msg = new StringBuilder();
+            msg.append("Unpacking ");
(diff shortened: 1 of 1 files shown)


slawekjaranowski:
Simillary  here title of PR not describe what we do .... 
Will be used in release notes, please look:
https://github.com/apache/maven-dependency-plugin/releases

slawekjaranowski:
also labels is used to categorize PR in release notes

elharo:
done

slawekjaranowski (approved):

PR #1638, opened 2026-05-29
[#1637] Fix dependency:add routing versioned deps into dependencyManagement

### LINKED ISSUE #1637: dependency:add puts versioned dependency in <dependencyManagement> instead of <dependencies> for quickstart/BOM-import POMs
### Affected version

`maven-dependency-plugin` (3.11.0), introduced with the `dependency:add` goal (#1599).

### Description

When running `dependency:add` with an **explicit version** on a project that has version-less dependencies managed by a **BOM import**, the dependency is incorrectly added to `<dependencyManagement>` of the **current** POM instead of `<dependencies>`.

This is exactly the layout produced by the standard `maven-archetype-quickstart` archetype: a `junit-bom` `import` in `<dependencyManagement>`, plus version-less `junit-jupiter-*` entries in `<dependencies>`.

### Steps to reproduce

Given a quickstart-style `pom.xml`:

```xml
<dependencyManagement>
  <dependencies>
    <dependency>
      <groupId>org.junit</groupId>
      <artifactId>junit-bom</artifactId>
      <version>5.11.0</version>
      <type>pom</type>
      <scope>import</scope>
    </dependency>
  </dependencies>
</dependencyManagement>

<dependencies>
  <dependency>
    <groupId>org.junit.jupiter</groupId>
    <artifactId>junit-jupiter-api</artifactId>
    <scope>test</scope>
  </dependency>
  <dependency>
    <groupId>org.junit.jupiter</groupId>
    <artifactId>junit-jupiter-params</artifactId>
    <scope>test</scope>
  </dependency>
</dependencies>
```

Run:

```bash
mvn dependency:add -Dgav=org.apache.commons:commons-lang3:3.20.0
```

### Expected

`commons-lang3:3.20.0` is added to `<dependencies>` (with its version).

### Actual

`commons-lang3` is added to the current POM's `<dependencyManagement>` instead. Because there is no separate `<dependencies>` entry, the dependency is not actually added to the project.

### Root cause

With the default `align=true`, `detectConventions()` flags the project as "uses managed dependencies" purely because the majority of `<dependencies>` are version-less:

```java
long totalDeps = getDependencyCount(editor, false);              // 2
List<String> depVersions = getDependencyVersions(editor, false); // [] (version-less)
if (totalDeps > 0 && depVersions.size() < totalDeps / 2.0) {     // true
    conv.useManaged = true;
}
```

But those dependencies are version-less only because their versions come from the imported BOM. There is no parent POM (`findManagedDepsPom()` returns `null`), so the cross-POM path is skipped and the managed dependency is written into the current POM's `<dependencyManagement>`.

This contradicts the documented behavior of the `align` parameter, whose Javadoc states the convention is to *"add managed dependency to **parent POM**"*.

### Proposed fix

The auto-detected `useManaged` convention should only apply when a separate parent POM exists to host the managed dependency. For a single/leaf POM, a versioned add should go to `<dependencies>`. Explicit `-Dmanaged=true` continues to target `<dependencyManagement>` as before.

brunoborges:
Fixes #1637

### Problem

`dependency:add` with an explicit version added the dependency to the current POM's `<dependencyManagement>` instead of `<dependencies>` when the project's existing dependencies are version-less because their versions come from a **BOM import** — the layout produced by `maven-archetype-quickstart`.

```bash
mvn dependency:add -Dgav=org.apache.commons:commons-lang3:3.20.0
```

With `align=true` (default), `detectConventions()` set `useManaged=true` purely because most `<dependencies>` were version-less. As there was no parent POM to host the managed entry (`findManagedDepsPom()` returned `null`), the dependency was written into the current POM's `<dependencyManagement>` — so it was never actually added as a usable dependency. This also contradicts the `align` Javadoc, which says the convention is to *"add managed dependency to **parent POM**"*.

### Fix

The auto-detected `useManaged` convention now only applies when a separate parent POM actually exists to host the managed dependency. For a single/leaf POM, a versioned add goes to `<dependencies>`. Explicit `-Dmanaged=true` is unchanged and still targets `<dependencyManagement>`.

### Testing

- New integration test `src/it/projects/add-dependency/bom-import` reproduces the quickstart/BOM-import scenario and asserts the dependency lands in `<dependencies>` and **not** in `<dependencyManagement>`.
- Existing `AddDependencyMojoTest` and `spotless:check` pass.
- Manually verified against the reported scenario: `commons-lang3:3.20.0` is now correctly added to `<dependencies>`.

### DIFF
--- a/src/it/projects/add-dependency/bom-import/pom.xml
+++ b/src/it/projects/add-dependency/bom-import/pom.xml
@@ -1,0 +1,54 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+  ~ Licensed to the Apache Software Foundation (ASF) under one
+  ~ or more contributor license agreements.  See the NOTICE file
+  ~ distributed with this work for additional information
+  ~ regarding copyright ownership.  The ASF licenses this file
+  ~ to you under the Apache License, Version 2.0 (the
+  ~ "License"); you may not use this file except in compliance
+  ~ with the License.  You may obtain a copy of the License at
+  ~
+  ~   http://www.apache.org/licenses/LICENSE-2.0
+  ~
+  ~ Unless required by applicable law or agreed to in writing,
+  ~ software distributed under the License is distributed on an
+  ~ "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+  ~ KIND, either express or implied.  See the License for the
+  ~ specific language governing permissions and limitations
+  ~ under the License.
+  -->
+
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>test</groupId>
+    <artifactId>add-dependency-bom-import</artifactId>
+    <version>1.0-SNAPSHOT</version>
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.junit</groupId>
+                <artifactId>junit-bom</artifactId>
+                <version>5.11.0</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <dependencies>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-api</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.junit.jupiter</groupId>
+            <artifactId>junit-jupiter-params</artifactId>
+            <scope>test</scope>
+        </dependency>
+    </dependencies>
+</project>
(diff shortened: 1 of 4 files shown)


brunoborges:
Hey folks, can I get a review on this fix? 

@cstamas @slawekjaranowski @slachiewicz

src/it/projects/add-dependency/bom-import/pom.xml
```diff
@@ -0,0 +1,54 @@
...
+
+    <dependencyManagement>
+        <dependencies>
+            <dependency>
+                <groupId>org.junit</groupId>
+                <artifactId>junit-bom</artifactId>
+                <version>5.11.0</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
```
slawekjaranowski, line 38: I hope we can use for such example any no existing artifacts, (like org.example:bom) as ddd mojo not require dependencies resolving, or we can add next mock artifact like a1 is.

When we merge it dependabot will try to updates it, and we will have next maintenance works on list

slawekjaranowski (changes requested):

src/it/projects/add-dependency/bom-import/pom.xml
brunoborges: @slawekjaranowski can we merge this one as is, and I will submit another PR changing all artifacts to non-existent examples.

slawekjaranowski (approved):

src/it/projects/add-dependency/bom-import/pom.xml
slawekjaranowski: ok, waiting for next PR 😄

PR #1666, opened 2026-08-01
Remove unreachable null check from BuildClasspathMojo

### LINKED ISSUE #1648: NPE in BuildClasspathMojo when getResolvedDependencies returns null
`BuildClasspathMojo.execute()` at line 199-205:

```java
Set<Artifact> artifacts = getResolvedDependencies(true);

if (artifacts == null || artifacts.isEmpty()) {
    getLog().info("No dependencies found.");
}

List<Artifact> artList = new ArrayList<>(artifacts);  // NPE if null
```

When `getResolvedDependencies()` returns `null`, the code logs "No dependencies found." but then falls through to `new ArrayList<>(null)`, which throws NullPointerException. There is no `return` statement after the null/empty check.

If `artifacts` is empty (not null), execution continues normally and produces an empty classpath string, which is harmless but inconsistent.

wilx:
`getResolvedDependencies(true)` supplies a non-null `Set`; projects with no resolved dependencies receive an empty set. Remove the misleading `artifacts == null` condition from `BuildClasspathMojo` while retaining the existing empty-set logging and empty classpath output.

This aligns the Mojo with the dependency resolver's effective contract. Because the null case cannot be produced through the supported Maven dependency-resolution path, no defensive conversion or null-specific test is needed.

Fixes #1648.

  This is not a behavioral change: the null branch is unreachable, and the existing empty-set behavior is unchanged.

### DIFF
--- a/src/main/java/org/apache/maven/plugins/dependency/fromDependencies/BuildClasspathMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/fromDependencies/BuildClasspathMojo.java
@@ -198,7 +198,7 @@ public class BuildClasspathMojo extends AbstractDependencyFilterMojo implements 
 
         Set<Artifact> artifacts = getResolvedDependencies(true);
 
-        if (artifacts == null || artifacts.isEmpty()) {
+        if (artifacts.isEmpty()) {
             getLog().info("No dependencies found.");
         }
 


src/main/java/org/apache/maven/plugins/dependency/fromDependencies/BuildClasspathMojo.java
```diff
@@ -198,7 +199,10 @@ protected void doExecute() throws MojoExecutionException {
 
         Set<Artifact> artifacts = getResolvedDependencies(true);
```
elharo, line 199: can this method return null? what's it's contract? and if it can return null, should it be changed to return an empty set instead?

elharo (changes requested):
needs a test

src/main/java/org/apache/maven/plugins/dependency/fromDependencies/BuildClasspathMojo.java
wilx: Not in the current Maven, I think. But I did not study all the versions.

elharo: Is the method marked nullable or non-nullable?

wilx: It is not marked.

elharo: I don't see an issue here. If the method doesn;t return null, it doesn't return null.

wilx: I have removed the check and reworded the commit.

wilx:
I have removed the null check entirely.

elharo (approved):

PR #1667, opened 2026-08-01
[MDEP-726] Resolve artifact item versions from transitive dependencies

### LINKED ISSUE #1240: [MDEP-726] dependency:copy finds missing versions in TRANSITIVE dependencies
**[Markus Karg](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=mkarg)** opened **[MDEP-726](https://issues.apache.org/jira/browse/MDEP-726?redirect=false)** and commented

In case someone wants to dependency:copy a transitive dependency, currently he needs to explicitly provide that artifact's version. This is a bit annoying, because each time a transitive dependency gets updated, the POM containing the copy-execution needs to get adjusted. It would be great, if version could get omitted, just like it already works with direct dependencies and managed dependencies.

---

**Affects:** 3.3.0

**Remote Links:**
- [GitHub Pull Request #109
](https://github.com/apache/maven-dependency-plugin/pull/109)

wilx:
## Summary

Allow `dependency:copy` and `dependency:unpack` artifact items to omit their version when the artifact is present only as a transitive project dependency.

The existing direct-dependency and dependency-management lookups remain authoritative. If neither supplies a version, the plugin lazily collects the requested compile, runtime, or test dependency graph through Maven Resolver and looks up the selected artifact version. An optional `dependencyScope` setting resolves cases where different classpaths select different versions; without it, conflicting selections are reported rather than guessed.

Artifact matching first uses the complete group ID, artifact ID, type, and classifier coordinates and then preserves the existing GA fallback. The resolved graphs are cached per scope in immutable GA-indexed maps. Existing relocation behavior is retained because the original model coordinates are checked before Resolver's post-relocation graph.

This implementation uses the Maven shared dependency graph APIs supported by both Maven 3.x and Maven 4.x.

Fixes #1240.

## Verification

- `mvn verify` — 415 tests passed, with one existing skip.
- `mvn -Prun-its verify` — all 103 integration projects passed.
- Focused `mdep-726-*`, `copy-relocation`, and `unpack-relocation` integration tests passed with:
  - Maven 3.6.3 on JDK 8
  - Maven 3.9.16 on JDK 21
  - Maven 4.0.0-rc-5 on JDK 21
- Spotless, Checkstyle, Apache RAT, dependency analysis, and `git diff --check` passed.

### DIFF
--- a/src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/AbstractFromConfigurationMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/AbstractFromConfigurationMojo.java
@@ -129,8 +129,8 @@ public abstract class AbstractFromConfigurationMojo extends AbstractDependencyMo
         super(session, buildContext, project);
         this.artifactHandlerManager = artifactHandlerManager;
         this.resolverUtil = resolverUtil;
-        this.dependencyVersionResolver =
-                new DependencyVersionResolver(session, project, new DefaultDependencyCollectorBuilder(repositorySystem));
+        this.dependencyVersionResolver = new DependencyVersionResolver(
+                session, project, new DefaultDependencyCollectorBuilder(repositorySystem));
     }
 
     abstract ArtifactItemFilter getMarkedArtifactFilter(ArtifactItem item);
--- a/src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/AbstractFromConfigurationMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/AbstractFromConfigurationMojo.java
@@ -37,7 +37,8 @@ import org.apache.maven.plugins.dependency.utils.ResolverUtil;
 import org.apache.maven.plugins.dependency.utils.filters.ArtifactItemFilter;
 import org.apache.maven.project.MavenProject;
 import org.apache.maven.shared.artifact.filter.collection.ArtifactFilterException;
-import org.apache.maven.shared.dependency.graph.DependencyCollectorBuilder;
+import org.apache.maven.shared.dependency.graph.internal.DefaultDependencyCollectorBuilder;
+import org.eclipse.aether.RepositorySystem;
 import org.eclipse.aether.RepositorySystemSession;
 import org.eclipse.aether.artifact.DefaultArtifact;
 import org.eclipse.aether.resolution.ArtifactResolutionException;
@@ -123,12 +124,13 @@ public abstract class AbstractFromConfigurationMojo extends AbstractDependencyMo
             BuildContext buildContext,
             MavenProject project,
             ArtifactHandlerManager artifactHandlerManager,
-            ResolverUtil resolverUtil,
-            DependencyCollectorBuilder dependencyCollectorBuilder) {
+            RepositorySystem repositorySystem,
+            ResolverUtil resolverUtil) {
         super(session, buildContext, project);
         this.artifactHandlerManager = artifactHandlerManager;
         this.resolverUtil = resolverUtil;
-        this.dependencyVersionResolver = new DependencyVersionResolver(session, project, dependencyCollectorBuilder);
+        this.dependencyVersionResolver =
+                new DependencyVersionResolver(session, project, new DefaultDependencyCollectorBuilder(repositorySystem));
     }
 
     abstract ArtifactItemFilter getMarkedArtifactFilter(ArtifactItem item);
--- a/src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/AbstractFromConfigurationMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/AbstractFromConfigurationMojo.java
@@ -21,7 +21,6 @@ package org.apache.maven.plugins.dependency.fromConfiguration;
 import java.io.File;
 import java.util.Collections;
 import java.util.List;
-import java.util.Objects;
 
 import org.apache.commons.lang3.StringUtils;
 import org.apache.maven.RepositoryUtils;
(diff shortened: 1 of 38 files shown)


src/it/projects/copy-relocation/pom.xml
```diff
@@ -52,7 +60,6 @@
                 <artifactItem>
                   <groupId>org.apache.maven.its.dependency</groupId>
                   <artifactId>ResourceArtifact-relocation</artifactId>
-                  <version>1.0</version>
```
elharo, line 55: we should probably keep the existing test as is and add new ones without the version

src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/AbstractFromConfigurationMojo.java
```diff
@@ -120,21 +115,21 @@ public abstract class AbstractFromConfigurationMojo extends AbstractDependencyMo
...
+    private final DependencyVersionResolver dependencyVersionResolver;
+
     protected AbstractFromConfigurationMojo(
             MavenSession session,
             BuildContext buildContext,
             MavenProject project,
             ArtifactHandlerManager artifactHandlerManager,
-            RepositorySystem repositorySystem,
-            ResolverUtil resolverUtil) {
+            ResolverUtil resolverUtil,
```
elharo: This incompatibly changes a published API

src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/ArtifactItem.java
```diff
@@ -53,6 +53,16 @@ public class ArtifactItem implements DependableCoordinate {
     @Parameter
     private String version = null;
 
+    /**
+     * Classpath scope from which to infer a missing version when the artifact is not found in direct dependencies or
+     * dependency management. Supported values are {@code compile}, {@code runtime}, and {@code test}. When omitted,
+     * all three dependency graphs are considered and must agree on the selected version.
```
elharo, line 58: what if they don't agree?

src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/ArtifactItem.java
```diff
@@ -53,6 +53,16 @@ public class ArtifactItem implements DependableCoordinate {
...
+     *
+     * @since 3.11.1
+     */
+    @Parameter
+    private String dependencyScope;
```
elharo, line 63: scope is different than the other fields. They're about the artifact itself. This is about how the artifact is used. This might be confusing artifacts and dependencies on artifacts. An artifact doesn't really have a scope.

elharo (changes requested):

src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/ArtifactItem.java
wilx: Then it is an error and the user has to deal with it by either specifying the `dependencyScope` or by giving explicit version`.

src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/ArtifactItem.java
wilx: See above. This is used to guide the dependency version resolution for the artifact if the version is ambiguous due to different version of the artifact in different scopes.

src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/AbstractFromConfigurationMojo.java
wilx: There is not a single use that is outside of this plugin that I could find in any Maven sources.

But I will try to preserve the signature.

wilx: I reverted the signature change.

src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/ArtifactItem.java
elharo: will this mean users now get errors on projects where they saw no errors before?

wilx: No. This entire patch is strictly improving the situation. Before the patch, you had to add the version explicitly. With the patch, you can omit the version and it will get resolved but it has to be unambiguous.

src/it/projects/copy-relocation/pom.xml
wilx: I reverted this and added separate tests.

wilx:
I have pushed the formatting fix.

wilx:
Rebased.

wilx:
Rebased.

wilx:
Ping.

src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/ClasspathScope.java
```diff
@@ -0,0 +1,59 @@
...
+    COMPILE,
+    RUNTIME,
+    TEST;
+
+    static ClasspathScope fromString(String value) {
+        try {
+            return valueOf(value.toUpperCase(Locale.ROOT));
+        } catch (IllegalArgumentException e) {
+            throw new IllegalArgumentException(
+                    "Unknown dependency scope '" + value + "'. Expected one of: compile, runtime, test.", e);
```
elharo: Probably not helpful to wrap the original exception here.

elharo (approved):

src/main/java/org/apache/maven/plugins/dependency/fromConfiguration/ClasspathScope.java
wilx: Do you mean you want to remove the `e` argument and ignore the original exception?

elharo: Yes, in this case the bug is really here in this method, not inside the Enum code, so there doesn't seem to be any benefit to wrapping the original exception. The stack trace will be cleaner and point more directly to the real issue if we just throw an exception here.

wilx: Fixed.

wilx:
Rebased.

PR #1669, opened 2026-08-02
[MDEP-870] Overwrite copied artifact POMs

### LINKED ISSUE #1370: [MDEP-870] artifact pom not overwritten
**[Jürgen](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=work_registries)** opened **[MDEP-870](https://issues.apache.org/jira/browse/MDEP-870?redirect=false)** and commented

the situation:

copy-dependencies, copyPom true, stripVersion true

no clean of target directory beforehand

the artifact jar + associated pom already in target directory (with different artifact version due to project branch switch)

./artifact.jar (1.0.0)
./artifact.pom (1.0.0)

what happens, copy-dependencies will copy the new artifact.jar to target directory, but not copy the associated pom, leaving the old (for a different version) pom

./artifact.jar (2.0.0, 1.0.0 was overwritten)
./artifact.pom (1.0.0, not overwritten!)

my gut feeling is, that 

```java
CopyDependenciesMojo.copyPoms() {
...
   if (!pomDestFile.exists()) copyFile
...
}
```

is the wrong check. the pom should be copied/overwritten, if it's artifact was copied/overwritten.

cf. https://github.com/apache/maven-dependency-plugin/blob/e52bc0248c00dbf5458a0ce080db260148dab4b9/src/main/java/org/apache/maven/plugins/dependency/fromDependencies/CopyDependenciesMojo.java#L286

same is true for current 3.6.0

---

**Affects:** 3.3.0, 3.6.0

wilx:
## Summary

When `dependency:copy-dependencies` copies an artifact, overwrite its associated POM as well. This prevents version-stripped output from containing a JAR from one version and a pre-existing POM from another version.

Artifacts skipped by the configured overwrite policy retain the existing MDEP-164 behavior: their POM is copied only when it is missing.

## Root cause

Copied and skipped artifacts shared the same `copyPoms` path, which unconditionally ignored an existing destination POM. As a result, an artifact selected for replacement could overwrite the versionless JAR while leaving its old versionless POM untouched.

Fixes #1370

## Verification

- The new regression test failed before the production change with `expected: <new pom> but was: <old pom>`.
- Maven 3: `mvn verify` — 412 tests, 0 failures/errors, 1 skipped.
- Maven 4 with JDK 21: `mvn verify` — 412 tests, 0 failures/errors, 1 skipped.
- Focused Invoker build: `mvn -DskipTests -Prun-its -Dinvoker.test=copy-dependencies verify` — 1 passed.
- Exact Commons Lang 3.12.0 to 3.14.0 reproducer: the resulting versionless JAR and POM checksums both match 3.14.0.

- [ ] You have run the integration tests successfully (`mvn -Prun-its verify`).

### DIFF
--- a/src/main/java/org/apache/maven/plugins/dependency/fromDependencies/CopyDependenciesMojo.java
+++ b/src/main/java/org/apache/maven/plugins/dependency/fromDependencies/CopyDependenciesMojo.java
@@ -170,9 +170,9 @@ public class CopyDependenciesMojo extends AbstractFromDependenciesMojo {
         }
 
         if (isCopyPom() && !useRepositoryLayout) {
-            copyPoms(getOutputDirectory(), artifacts, this.stripVersion);
-            copyPoms(getOutputDirectory(), skippedArtifacts, this.stripVersion, this.stripClassifier);
+            copyPoms(getOutputDirectory(), artifacts, this.stripVersion, this.stripClassifier, true);
             // Artifacts that already exist may not yet have poms
+            copyPoms(getOutputDirectory(), skippedArtifacts, this.stripVersion, this.stripClassifier, false);
         }
     }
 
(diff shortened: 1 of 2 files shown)


wilx:
```
Suppressed: org.eclipse.aether.transfer.ArtifactTransferException: Could not transfer artifact org.ow2.asm:asm-util:pom:9.9 from/to central (https://repo.maven.apache.org/maven2): HTTP connect timed out
```

Some sort of infrastructure issue.

slachiewicz (approved):

PR #1670, opened 2026-08-02
[MDEP-425] List plugin repositories

### LINKED ISSUE #868: [MDEP-425] dependency:list-repositories ignores pluginRepository (only lists repository)
**[Michael Vorburger](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=vorburger)** opened **[MDEP-425](https://issues.apache.org/jira/browse/MDEP-425?redirect=false)** and commented

http://maven.apache.org/plugins/maven-dependency-plugin/list-repositories-mojo.html can be very useful! But unless we've missed something, it seems to only gather & list all \<repositories>\<repository> but doesn't know about / ignores any \<pluginRepositories>\<pluginRepository> ?

What I'm suggesting is more that something like 'mvn org.apache.maven.plugins:maven-dependency-plugin:2.8:list-repositories' also lists plugin repositories (with an indication of which is a plugin vs. a dependency plug-in?) than a new list-plugin-repositories goal (just because I think in practice one would typically want to see both).

---

**Remote Links:**
- [GitHub Pull Request #113
](https://github.com/apache/maven-plugins/pull/113)
- [GitHub Pull Request #3
](https://github.com/apache/maven-dependency-plugin/pull/3)
- [GitHub Pull Request #51
](https://github.com/apache/maven-dependency-plugin/pull/51)

wilx:
Fixes #868.

The `list-repositories` goal currently seeds dependency collection from the project's dependency repositories and reports repositories discovered in that dependency graph. Plugin repositories are available separately from the effective Maven project, so they never enter that collection and are omitted from the output.

This change reads the effective plugin repositories directly from `MavenProject`, then prints them in a distinct, mirror-aware section. The existing integration test now configures a plugin repository and verifies that both dependency and plugin repository sections are reported.

Validation:

- Confirmed the new integration-test assertions fail before the runtime change.
- `mvn verify` with Maven 3.9.16 and JDK 17.
- `mvn -Prun-its verify` with Maven 3.9.16 and JDK 17: 410 tests passed with one existing skip; all 99 Invoker builds passed.
- Focused `list-repositories` IT with Maven 4.0.0-rc-5 and JDK 17.

### DIFF
--- a/src/it/projects/list-repositories/pom.xml
+++ b/src/it/projects/list-repositories/pom.xml
@@ -43,6 +43,13 @@
     </repository>
   </repositories>
 
+  <pluginRepositories>
+    <pluginRepository>
+      <id>fake-plugin-repository</id>
+      <url>http://localhost:3456</url>
+    </pluginRepository>
+  </pluginRepositories>
+
   <dependencies>
     <dependency>
       <groupId>org.apache.maven</groupId>
--- a/src/it/projects/list-repositories/verify.groovy
+++ b/src/it/projects/list-repositories/verify.groovy
@@ -23,6 +23,8 @@ assert file.exists()
 String buildLog = file.getText( "UTF-8" )
 assert buildLog.contains( 'Project remote repositories used by this build:')
 assert buildLog.contains( '* fake-remote-repository (http://localhost:2345, default, releases+snapshots)')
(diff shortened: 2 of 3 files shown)


elharo (commented):
I'm not 100% sold on this idea. +0

slachiewicz (approved):

PR #1671, opened 2026-08-02
[MDEP-927] Make dependency:tree excludes prune subtrees

### LINKED ISSUE #1334: [MDEP-829] excludes filter should function like exclusions
**[Delany](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=delany)** opened **[MDEP-829](https://issues.apache.org/jira/browse/MDEP-829?redirect=false)** and commented

The -Dincludes parameter causes the tree goal to filter the resultant dependency tree: it returns matching dependencies {**}and their parents{**}.

The -Dexcludes parameter uses the same logic, but returns non-matching dependencies {**}and their parents{**}. Why would anyone want this? Rather, excludes should remove matching dependencies from the initial dependency model (like dependency exclusions do), not just the resultant tree.

Include is demonstrated at https://maven.apache.org/plugins/maven-dependency-plugin/examples/filtering-the-dependency-tree.html, but not exclude.

If run the given exclude command

```java
mvn dependency:tree -Dexcludes=org.codehaus.plexus
```

on the following tree

```
[INFO] org.apache.maven.plugins:maven-dependency-plugin:maven-plugin:2.0-alpha-5-SNAPSHOT
[INFO] \- org.apache.maven.doxia:doxia-site-renderer:jar:1.0-alpha-8:compile
[INFO]    \- org.codehaus.plexus:plexus-velocity:jar:1.1.3:compile
[INFO]       \- velocity:velocity:jar:1.4:compile
```

the output is the same, since the velocity dependency needs to show its parent in the tree. This is pointless. Why would I only want to exclude the "leaves" of the tree?

---

**Affects:** 3.3.0

### LINKED ISSUE #1427: [MDEP-927] dependency:tree excludes parameter does not work
**[Seva Popov](https://issues.apache.org/jira/secure/ViewProfile.jspa?name=sevapopov)** opened **[MDEP-927](https://issues.apache.org/jira/browse/MDEP-927?redirect=false)** and commented

The dependency:tree excludes parameter does not work in any plugin version available for download.

Ex.:

mvn org.apache.maven.plugins:maven-dependency-plugin:3.6.1:tree -s ~/.m2/settings.xml -Dexcludes=javax:javaee-api:jar:8.0.1  | grep javax.javaee

[INFO] +- javax:javaee-api:jar:8.0.1:provided

According to the documentation the dependency should have been excluded - https://maven.apache.org/plugins/maven-dependency-plugin/tree-mojo.html#excludes

---

**Affects:** 3.6.1

wilx:
## Summary

- make `dependency:tree` prune a dependency matching `excludes` together with its complete subtree
- apply exclusions before includes, so an include cannot restore a node beneath an excluded subtree
- retain ancestor paths for included nodes that remain after exclusion pruning
- document the behavior and add unit and integration coverage for pruning, sibling traversal, and combined filters

## Root cause

The existing implementation combined `includes` and `excludes` into one node filter and applied it through `FilteringDependencyNodeVisitor`. A rejected node was omitted from the serialized output, but its children were still traversed. When includes were also present, the ancestor-or-self pass could restore paths through nodes that had matched an exclude pattern.

This change introduces a pruning visitor. A rejected node returns `false` from `visit`, preventing traversal of its children, while its matching `endVisit` returns `true` so traversal continues with the node's next sibling. The visitor remembers the acceptance decision made at the start of the visit so the delegate receives balanced `visit`/`endVisit` callbacks without evaluating the filter twice.

## Semantics and compatibility

This intentionally changes the output-filtering semantics of `dependency:tree`:

- `excludes` is a structural pruning operation. A matching node and every node below it are absent from the serialized tree.
- `includes` is a positive filter over the tree that remains after pruning. It selects matching dependencies and retains the ancestor paths needed to reach them.
- exclusions take precedence; includes cannot override them.
- this affects only serialization of the dependency tree and does not change dependency collection or resolution.

As discussed in #1334, the current parameter names do not describe these distinct roles especially well. Names such as `filter` for the path-preserving positive selection and `exclusions` for subtree pruning would communicate the behavior more clearly. This PR retains the established `includes` and `excludes` names for CLI and plugin-configuration compatibility. Introducing aliases, deprecating the existing names, or renaming them should be considered separately as an API migration.

## Validation

- `mvn spotless:apply`
- Maven 3: `mvn verify` — 412 tests passed, 1 skipped
- Maven 3: `mvn -Prun-its verify` — 99 integration-test projects passed
- Maven 4: `mvn verify` — 412 tests passed, 1 skipped
- Maven 4: `mvn -Prun-its verify` — 98 integration-test projects passed, 1 skipped due to its Maven-version condition
- Spring Web Services 4.0.11 reproducer with the snapshot plugin, both normal and `-Dverbose=true`: excluding `org.glassfish.jaxb:jaxb-runtime` also removes `jaxb-core`, `txw2`, and `istack-commons-runtime`

Fixes #1427

Fixes #1334

### DIFF
--- a/src/it/projects/tree-excluded/expected-included-survives.txt
+++ b/src/it/projects/tree-excluded/expected-included-survives.txt
@@ -1,0 +1,4 @@
+org.apache.maven.its.dependency:test:jar:1.0-SNAPSHOT
+\- org.apache.maven:maven-project:jar:2.0.6:compile
+   \- org.apache.maven:maven-artifact-manager:jar:2.0.6:compile
+      \- org.apache.maven.wagon:wagon-provider-api:jar:1.0-beta-2:compile
--- a/src/it/projects/tree-excluded/expected-v4.txt
+++ b/src/it/projects/tree-excluded/expected-v4.txt
@@ -7,7 +7,4 @@ org.apache.maven.its.dependency:test:jar:1.0-SNAPSHOT
    |  +- org.apache.maven:maven-repository-metadata:jar:2.0.6:compile
    |  \- org.apache.maven.wagon:wagon-provider-api:jar:1.0-beta-2:compile
    +- org.apache.maven:maven-plugin-registry:jar:2.0.6:compile
-   +- org.apache.maven:maven-artifact:jar:2.0.6:compile
-   \- org.codehaus.plexus:plexus-container-default:jar:1.0-alpha-9-stable-1:compile
-      +- junit:junit:jar:3.8.1:compile
-      \- classworlds:classworlds:jar:1.1:compile
+   \- org.apache.maven:maven-artifact:jar:2.0.6:compile
--- a/src/it/projects/tree-excluded/expected.txt
+++ b/src/it/projects/tree-excluded/expected.txt
@@ -7,7 +7,4 @@ org.apache.maven.its.dependency:test:jar:1.0-SNAPSHOT
    |  +- org.apache.maven:maven-repository-metadata:jar:2.0.6:compile
    |  \- org.apache.maven.wagon:wagon-provider-api:jar:1.0-beta-2:compile
    +- org.apache.maven:maven-plugin-registry:jar:2.0.6:compile
-   +- org.apache.maven:maven-artifact:jar:2.0.6:compile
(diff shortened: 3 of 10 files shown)


elharo (approved):

wilx:
Rebased.

PR #1673, opened 2026-08-05
Clarify pull_request_template.md

elharo:


slawekjaranowski:
We have such template in many Maven projects, should be similar in all.
Do you want to update all projects?

Please also look at 
- https://github.com/apache/maven-gh-actions-shared/issues/165

elharo:
Eventually, maybe, but we do not have a monorepo or genuinely shared files, just copy and paste, so every repo needs its own PR.

Issue #1682, opened 2026-08-10
render-dependencies breaks on Maven 3.6.3, the declared prerequisite


slachiewicz:
The POM declares

```xml
<prerequisites>
  <maven>3.6.3</maven>
</prerequisites>
```

but builds against `<mavenVersion>3.9.16</mavenVersion>`, and nothing in the build or in CI ever
checks the plugin against the version it claims to support. The two have drifted, and the drift
is already in released artifacts.

### The defect

`RenderDependenciesMojo` line 241 uses the single-argument `MojoExecutionException(Throwable)`
constructor:

```java
} catch (final IOException e) {
    throw new MojoExecutionException(e);
}
```

That constructor was added in Maven 3.9.0. It does not exist in 3.6.3:

```
$ javap -classpath maven-plugin-api-3.6.3.jar org.apache.maven.plugin.MojoExecutionException
  MojoExecutionException(java.lang.Object, java.lang.String, java.lang.String);
  MojoExecutionException(java.lang.String, java.lang.Exception);
  MojoExecutionException(java.lang.String, java.lang.Throwable);
  MojoExecutionException(java.lang.String);

$ javap -classpath maven-plugin-api-3.9.16.jar org.apache.maven.plugin.MojoExecutionException
  ... the four above, plus:
  MojoExecutionException(java.lang.Throwable);
```

The reference is in the shipped bytecode, not just the source — from the released
`maven-dependency-plugin-3.11.0.jar`:

```
$ javap -c -p org.apache.maven.plugins.dependency.fromDependencies.RenderDependenciesMojo
  34: invokespecial #298  // Method org/apache/maven/plugin/MojoExecutionException."<init>":(Ljava/lang/Throwable;)V
```

So on Maven 3.6.3 this throws `NoSuchMethodError` rather than the intended
`MojoExecutionException`. The path is narrow — it is the error branch, reached when
`Files.createDirectories` fails while writing the output file — but it turns a diagnosable
build failure into a linkage error.

`RenderDependenciesMojo` was added in 23186d44 (#1523) and is present in released
**3.9.0, 3.10.0 and 3.11.0**.

### Reproducing

```
mvn -DmavenVersion=3.6.3 clean test-compile
```

```
[ERROR] .../RenderDependenciesMojo.java:[241,50] incompatible types:
        java.io.IOException cannot be converted to java.lang.String
```

(javac resolves the call to `MojoExecutionException(String)` once the `Throwable` overload is
[...]

PR #1680, opened 2026-08-09
[MDEP-858] Complete Maven Artifact Transfer removal after #1677

wilx:
> [!IMPORTANT]
> This PR is stacked on top of #1677 and is the stacked alternative to #1668. Until #1677 is merged, GitHub's diff also includes its two commits. After #1677 is merged, this PR should contain only the five follow-up commits described below. If this stacked approach is preferred, #1668 can be closed.

## Summary

Complete the Maven Artifact Transfer removal after #1677:

- migrate repository-layout installation in `dependency:copy-dependencies` to Maven Resolver while preserving local-repository metadata, copied POMs, snapshot base versions, and repository content type;
- migrate `dependency:purge-local-repository` resolution while preserving filtering, fallback resolution, classifiers, and aggregated failure reporting;
- remove the now-unused `maven-artifact-transfer` dependency and `DependableCoordinate` implementation;
- preserve `dependency:get`'s historical unconditional refresh policy for explicitly supplied repositories while leaving other `ResolverUtil` callers session-controlled;
- preserve separately configured packaging and classifier values for short artifact coordinates, with explicit coordinate segments retaining precedence.

The repository parsing, strict artifact-coordinate validation, and `dependency:get` Resolver migration come from #1677.

Fixes #1355

## Verification

- Maven 3/JDK 21: `nice mvn clean verify` — 434 tests, 0 failures/errors, 1 skipped; dependency analysis clean
- Maven 3/JDK 21: `nice mvn clean -Prun-its verify` — 99 Invoker projects passed, 0 failed/errors/skipped
- Maven 4/JDK 21: focused `TestGetMojo`, `ResolverUtilTest`, and `TestCopyDependenciesMojo2` clean run — 43 tests, 0 failures/errors/skips
- Maven 3/JDK 21: focused `get-*` Invoker run — 4 passed, 0 failed/errors/skipped
- `spotless:check` passed on every follow-up commit revision; `git diff --check` passed on the final head

Following this checklist to help us incorporate your
contribution quickly and easily:

- [x] Your pull request should address just one issue, without pulling in other changes.
- [x] Write a pull request description that is detailed enough to understand what the pull request does, how, and why.
- [x] Each commit in the pull request should have a meaningful subject line and body.
  Note that commits might be squashed by a maintainer on merge.
- [x] Write unit tests that match behavioral changes, where the tests fail if the changes to the runtime are not applied.
  This may not always be possible but is a best-practice.
- [x] Run `mvn verify` to make sure basic checks pass.
[...]

PR #1683, opened 2026-08-10
Verify the plugin still compiles against its declared prerequisite

slachiewicz:
Adds a CI job that compiles main and test sources against the Maven version
`<prerequisites>` declares, so the plugin's supported baseline is actually tested.

```yaml
  baseline:
    name: Baseline (Maven 3.6.3 API)
    ...
      - run: mvn -B --no-transfer-progress -DmavenVersion=3.6.3 clean test-compile
```

`test-compile` rather than `verify` on purpose: API drift is a compile-time property, so
compiling both source roots is sufficient, and it avoids the cost of a second full test run.
It builds with the normal CI Maven — only the `${mavenVersion}` dependency stack moves — so
the inherited `requireMavenVersion` (3.9) is unaffected.

`clean` is load-bearing. Without it the compiler reports "Nothing to compile - all classes are
up to date" against a `target/` left by an earlier step, and the job passes green having
compiled nothing against the baseline.

Rebased onto master now that #1684 has landed, so the job is green. It was red before that fix,
which was the point — see #1682.

### Why the existing checks missed the drift

- The verify matrix runs `3.10.0-rc-1` and `4.0.0-rc-6`; 3.6.3 is never built.
- `requireMavenVersion` constrains the Maven that *builds* the project, not the API level it
  compiles against.
- A class-level linkage check (`jdeps -verbose:class`) reports it clean —
  `MojoExecutionException` exists in both versions and only the constructor differs. Catching it
  needs a member-level check, which is what compiling gives you.

### Known gap, deliberately not closed here

`<resolverVersion>` moves independently of `<mavenVersion>`, and `maven-resolver-api` is
`provided` scope. So under Maven 3.6.3 the plugin runs against the resolver 1.4.1 that Maven
ships, while this job still compiles it against 1.9.25 — the same drift class, still unguarded.

Adding `-DresolverVersion=1.4.1` costs nothing today: both master and #1680 compile clean
against it. Left out to keep this PR to a single change; happy to fold it in if preferred.

<sub>Drafted with Claude — please verify</sub>

### DIFF
--- a/.github/workflows/maven-verify.yml
+++ b/.github/workflows/maven-verify.yml
@@ -27,3 +27,26 @@ jobs:
     uses: apache/maven-gh-actions-shared/.github/workflows/maven-verify.yml@v5
     with:
       maven4-enabled: true
+
+  # The verify matrix runs current Maven only, so nothing checks the plugin against the
+  # version <prerequisites> claims to support. Compiling main and test sources against that
+  # baseline catches API drift without paying for a full test run.
+  baseline:
+    name: Baseline (Maven 3.6.3 API)
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@3d3c42e5aac5ba805825da76410c181273ba90b1 # v7.0.1
+
+      - name: Set up JDK
+        uses: actions/setup-java@b6effb05e454b25005698d916606bdc6ffcbf961 # v5.7.0
+        with:
+          java-version: '21'
+          distribution: 'zulu'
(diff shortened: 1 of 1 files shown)


slachiewicz:
Taking the first option: the fix is #1684, separately labelled so it lands in the release notes as a bug rather than under build. This job goes green once that merges; keeping it draft until then.

Checked that neither PR conflicts with #1680 — the file sets are disjoint, and #1680 compiles clean against the 3.6.3 baseline, so this job will not turn it red.

